PR #6, opened 2015-10-28
A validator for the request entity

cmrd-senya:


lib/diaspora_federation/validators/request_validator.rb
```diff
@@ -0,0 +1,10 @@
+module DiasporaFederation
+  module Validators
+    class RequestValidator < Validation::Validator
+      include Validation
+
+      rule :sender_id, [:not_nil, :diaspora_id]
```
diaspora-code-review: Use %i or %I for array of symbols.

lib/diaspora_federation/validators/request_validator.rb
```diff
@@ -0,0 +1,10 @@
...
+      rule :recipient_id, [:not_nil, :diaspora_id]
```
diaspora-code-review: Use %i or %I for array of symbols.

PR #9, opened 2015-11-05
DRY the specs more 

cmrd-senya:
by introducing "a property with data-types restriction" shared behavior


cmrd-senya:
This is superseded by #10.

PR #12, opened 2015-11-11
Fixes that in the XmlPaylod rspec ".entity_class_name" was mistakenly put inside the ".unpack" section

cmrd-senya:


SuperTux88:
thanks

PR #14, opened 2015-11-13
Introduce the "Relayable" entity

cmrd-senya:
Introduce the "Relayable" entity, which encapsulates common properties of entities that are inherited from Relayable class in the Diaspora source code - "parent_guid", "parent_author_signature", "author_signature".

### DIFF
--- a/lib/diaspora_federation/validators/relayable_validator.rb
+++ b/lib/diaspora_federation/validators/relayable_validator.rb
@@ -1,0 +1,11 @@
+module DiasporaFederation
+  module Validators
+    module RelayableValidator
+      def self.included(model)
+        model.class_eval do
+          rule :parent_guid, :guid
+        end
+      end
+    end
+  end
+end
--- a/spec/lib/diaspora_federation/validators/comment_validator_spec.rb
+++ b/spec/lib/diaspora_federation/validators/comment_validator_spec.rb
@@ -4,22 +4,16 @@ module DiasporaFederation
 
     it_behaves_like "a common validator"
 
+    it_behaves_like "a relayable validator"
+
     it_behaves_like "a diaspora id validator" do
       let(:property) { :diaspora_id }
       let(:mandatory) { true }
     end
 
-    %i(guid parent_guid).each do |prop|
+    describe "#guid" do
       it_behaves_like "a guid validator" do
-        let(:property) { prop }
-      end
-    end
-
-    %i(author_signature parent_author_signature).each do |prop|
-      describe "##{prop}" do
-        it_behaves_like "a property that mustn't be empty" do
-          let(:property) { prop }
-        end
+        let(:property) { :guid }
       end
     end
 
--- a/lib/diaspora_federation/entities.rb
+++ b/lib/diaspora_federation/entities.rb
@@ -8,6 +8,7 @@ module DiasporaFederation
   end
 end
 
+require "diaspora_federation/entities/relayable"
 require "diaspora_federation/entities/profile"
 require "diaspora_federation/entities/person"
 require "diaspora_federation/entities/location"
--- a/lib/diaspora_federation/entities/comment.rb
+++ b/lib/diaspora_federation/entities/comment.rb
@@ -2,9 +2,7 @@ module DiasporaFederation
   module Entities
     class Comment < Entity
       property :guid
-      property :parent_guid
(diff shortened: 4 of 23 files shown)


spec/lib/diaspora_federation/validators/comment_validator_spec.rb
```diff
@@ -15,14 +15,6 @@ module DiasporaFederation
...
     end
 
-    %i(author_signature parent_author_signature).each do |prop|
-      describe "##{prop}" do
-        it_behaves_like "a property that mustn't be empty" do
-          let(:property) { prop }
-        end
-      end
-    end
-
```
SuperTux88: Please add a `it_behaves_like "a relayable validator"` where you still test this.

lib/diaspora_federation/validators/relayable_validator.rb
```diff
@@ -0,0 +1,11 @@
+module DiasporaFederation
+  module Validators
+    module RelayableValidator
+      def self.included(model)
+        model.class_eval do
+          rule :parent_guid, :guid
```
SuperTux88: what about the `parent_author_signature` and `author_signature`?

cmrd-senya: The only applied rule was "not_empty", which actually isn't true. There are [no restriction](https://github.com/diaspora/diaspora/blob/develop/db/schema.rb#L170-L171) in DB schema and ```parent_author_signature``` is actually nil in [some](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/relayable.rb#L106-L109) circumstances.

spec/lib/diaspora_federation/validators/comment_validator_spec.rb
cmrd-senya: Done

cmrd-senya: But for ```parent_guid``` only, again.

lib/diaspora_federation/validators/relayable_validator.rb
SuperTux88: But I think at least `author_signature` should be there or the signature-verification will fail?

I don't know where the `parent_author_signature` is used and what happens if it is empty.

cmrd-senya: The code in Diaspora is not very clear, I think I should test this part more close to be sure.

cmrd-senya: Actually my tests show an opposite thing: parent_author_signature is mandatory and author_signature is optional. Here is logs:
[Both parameters are empty](https://gist.github.com/cmrd-senya/874abcaf75254e39d7cc#file-both-empty-log-L14).
[Both parameters are filled](https://gist.github.com/cmrd-senya/3212e954eec9801424da#file-both-filled-log-L10).
[Only author_signature is empty](https://gist.github.com/cmrd-senya/cd5669b86b598cd9554b#file-author_signature-empty-log-L10).
[Only parent_author_signature is empty](https://gist.github.com/cmrd-senya/d9b7bcad9c583bed1888#file-parent_author_signature-empty-log-L10).

Maybe that means a bug in Diaspora?

cmrd-senya: It looks like ```author_signature``` never gets checked really. The only thing that works for the relayable ancestors is ```parent_author_signature```.

cmrd-senya: @jhass, could you please check this outcomes? If it is a bug, we need to find out, how it is supposed to work.

jhass: I think you're probably deeper into the related code atm that I am. Did you compare behaviour between

* parent author is local
* parent author is remote
* author is local
* author is remote (received object as pod hosting the parent author)

cmrd-senya: I only ran test for a case when author and parent author is remote. But I don't see any place in the source code where ```author_signature``` check affects execution flow at all.

Actually, I can write some integration tests for this, if I know what is the idea behind it.

SuperTux88:
I merged this now, but we need to have a look at the signature-stuff again. But I think you can do this in #13.

PR #16, opened 2015-11-23
Introduce diaspora_federation-tests gem.

cmrd-senya:
ref [6539](https://github.com/diaspora/diaspora/pull/6539)


diaspora_federation-tests.gemspec
```diff
@@ -0,0 +1,24 @@
...
+require "diaspora_federation/version"
+
+# Describe your gem and declare its dependencies:
+Gem::Specification.new do |s|
+  s.name        = "diaspora_federation-tests"
+  s.version     = DiasporaFederation::VERSION
+  s.authors     = ["Benjamin Neff"]
+  s.email       = ["benjamin@coding4.coffee"]
+  s.homepage    = "https://github.com/SuperTux88/diaspora_federation"
+  s.summary     = "diaspora* federation library"
```
SuperTux88: "diaspora* federation test utils" (or something like this)

diaspora_federation-tests.gemspec
```diff
@@ -0,0 +1,24 @@
...
+  s.description = "This gem provides some supplimentary code (factory definitions), that"\
+                  "help to build tests for users of the diaspora_federation gem."
+  s.license     = "AGPL 3.0 - http://www.gnu.org/licenses/agpl-3.0.html"
+
+  s.files       = Dir["lib/diaspora_federation/factories.rb"]
```
SuperTux88: maybe move it to a "test" subfolder?

diaspora_federation-tests.gemspec
```diff
@@ -0,0 +1,24 @@
...
+
+  s.required_ruby_version = "~> 2.0"
+
+  s.add_dependency "diaspora_federation", DiasporaFederation::VERSION
+  s.add_dependency "factory_girl_rails", "~> 4.5.0"
```
SuperTux88: do we need "factory_girl_rails" here or is "factory_girl" enough?

diaspora_federation-tests.gemspec
SuperTux88: or maybe "tests" because the gem is named "...-tests"

SuperTux88:
something like this was also my idea, but you were faster ;)

diaspora_federation-tests.gemspec
cmrd-senya: Having test and tests folders at the same time is confusing

SuperTux88: then maybe name the gem "diaspora_federation-test"?

I don't mean the "test" folder at toplevel, I mean "lib/diaspora_federation/test/factories.rb". You should also add a "lib/diaspora_federation/test.rb" as entry point for the gem. (see "lib/diaspora_federation/rails.rb" for the -rails gem)

SuperTux88: see: http://guides.rubygems.org/name-your-gem/

PR #17, opened 2015-11-25
Documentation added

cmrd-senya:
Here are some documentation updates.
It's not finished yet, here I'll post further updates.

### DIFF
--- a/lib/diaspora_federation/entities/like.rb
+++ b/lib/diaspora_federation/entities/like.rb
@@ -1,10 +1,31 @@
 module DiasporaFederation
   module Entities
+    # this entity represents a like to some kind of post (e.g. status message)
+    #
+    # @see Validators::LikeValidator
     class Like < Entity
+      # @!attribute [r] positive
+      #   If true set a like, if false, set a dislike (dislikes are currently not
+      #   implemented in the Diaspora's frontend).
+      #   @return [Boolean] is it a like or a dislike
       property :positive
+
+      # @!attribute [r] guid
+      #   @see HCard#guid
+      #   @return [String] guid
       property :guid
+
+      # @!attribute [r] target_type
+      #   a string describing a type of the target
+      #   @return [String] target type
       property :target_type
+
       include Relayable
+
+      # @!attribute [r] diaspora_id
+      #   The diaspora ID of the person who posts a like
+      #   @see Person#diaspora_id
+      #   @return [String] diaspora ID
       property :diaspora_id, xml_name: :diaspora_handle
     end
   end
--- a/lib/diaspora_federation/entities/relayable.rb
+++ b/lib/diaspora_federation/entities/relayable.rb
@@ -1,10 +1,32 @@
 module DiasporaFederation
   module Entities
+    # this is a module that defines common properties for relayable entities
+    # which include Like, Comment, Participation, Message, etc. Each relayable
+    # has a parent, identified by guid. Relayables also are signed and signing/verificating
+    # logic is embedded into Salmon XML processing code.
     module Relayable
+      # on inclusion of this module the required properties for a relayable are added to the object that includes it
       def self.included(model)
         model.class_eval do
+          # @!attribute [r] parent_guid
+          #   @see HCard#guid
+          #   @return [String] parent guid
           property :parent_guid
+
+          # @!attribute [r] parent_author_signature
+          #   Contains a signature of the entity using the private key of the author of a parent post
+          #   This signature is required only when federation from upstream (parent) post author to
+          #   downstream subscribers. This is the case when the parent author has to resend a relayable
+          #   received from one of his subscribers to all others.
+          #
+          #   @return [String] parent author signature
           property :parent_author_signature, default: nil
+
+          # @!attribute [r] author_signature
+          #   Contains a signature of the entity using the private key of the author of a post itself.
+          #   The presence of this signature is mandatory. Without it the entity won't be accepted by
+          #   a target pod.
+          #   @return [String] author signature
           property :author_signature, default: nil
         end
       end
@@ -20,6 +42,7 @@ module DiasporaFederation
         xml
       end
 
+      # Exception raised when verify_signatures fails to verify signatures (signatures are wrong)
       class SignatureVerificationFailed < ArgumentError
       end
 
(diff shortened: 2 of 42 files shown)


lib/diaspora_federation/entities/relayable.rb
```diff
@@ -41,6 +58,11 @@ def self.verify_signatures(data)
         end
       end
 
+      # Adds signatures to a given hash with the keys of the author and the parent
+      # if the signatures are not in the hash yet and if the keys are available.
+      #
+      # @param [Hash] data hash given for a signing
+      # @return [Hash] reference to the input hash
```
diaspora-code-review: Assignment Branch Condition size for verify_signatures is too high. [21.68/20]

cmrd-senya:
I've added documentation to all the objects I had introduced. Only two methods left reported by yardoc:
```
Files:          81
Modules:        15 (    0 undocumented)
Classes:        84 (    0 undocumented)
Constants:      22 (    0 undocumented)
Methods:       211 (    2 undocumented)
 99.40% documented

Undocumented Objects:

(in file: lib/diaspora_federation/salmon/slap.rb)
DiasporaFederation::Salmon::Slap.build_xml
DiasporaFederation::Salmon::Slap#add_magic_env_from_doc
```

I'm unsure what to write.

I also noticed that some method are not reported by yardoc, though they don't have documentation defined.

SuperTux88:
I can add the missing ones, and I think yardoc only shows public methods.

lib/diaspora_federation/entities/relayable_retraction.rb
```diff
@@ -1,10 +1,33 @@
...
       property :target_type
+
+      # @!attribute [r] sender_id
+      #   The diaspora ID of the person who deletes a post
+      #   @see Person#diaspora_id
+      #   @return [String] diaspora ID
       property :sender_id, xml_name: :sender_handle
+
+      # @!attribute [r] target_author_signature
+      #   Contains a signature of the entity using the private key of the author of a post itself
```
SuperTux88: I think this is the signature of the author of the comment/like/whatever?

Do you know what happens if the author of the parent-post deletes a comment? (the author if a post has the permissions to moderate the comments). who checks this signature and when is it needed?

I think we should document this things too, if we know it :)

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -1,10 +1,26 @@
...
+      # on inclusion of this module the required properties for a relayable are added to the object that includes it
       def self.included(model)
         model.class_eval do
+          # @!attribute [r] parent_guid
+          #   @see HCard#guid
+          #   @return [String] parent guid
           property :parent_guid
+
+          # @!attribute [r] parent_author_signature
+          #   Contains a signature of the entity using the private key of the author of a parent post
```
SuperTux88: Can you add the info here, when which signature needs to be verified?

lib/diaspora_federation/entities/signed_retraction.rb
```diff
@@ -1,9 +1,28 @@
...
       property :target_type
+
+      # @!attribute [r] sender_id
+      #   The diaspora ID of the person who deletes a post
+      #   @see Person#diaspora_id
+      #   @return [String] diaspora ID
       property :sender_id, xml_name: :sender_handle
+
+      # @!attribute [r] author_signature
+      #   Contains a signature of the entity using the private key of the author of a post
```
SuperTux88: when does a retraction need to be signed? (SignedRetraction vs Retraction)

lib/diaspora_federation/salmon/aes.rb
```diff
@@ -17,7 +17,6 @@ def self.generate_key_and_iv
       # @param [String] data plain input
       # @param [String] key AES key
       # @param [String] iv AES initialization vector
-      # @param [Hash] key_and_iv { key: "...", iv: "..." }
```
SuperTux88: thanks, nice catch, refactorings ... ;)

SuperTux88:
I have added some questions, but if you don't know the answers I have maybe some time at the weekend to do some research :)

lib/diaspora_federation/entities/signed_retraction.rb
cmrd-senya: The retractions are issued from [here](https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L377-L383).

RelayableRetraction is used for relayables, SignedRetraction is used for posts, and Retraction is used for everything else, which is ATM includes only photo.

lib/diaspora_federation/entities/relayable_retraction.rb
cmrd-senya: Yes, it is the signature of the relayable, I must reword it.

It works pretty much the same way the relayable federation works.

When author of the parent post deletes the comment this signature [is not checked](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/federated/relayable_retraction.rb#L51-L53).

cmrd-senya:
I added some answers on your questions. I added it as a different commit so it's easier to see the changes.

cmrd-senya:
Actually, it seems to me that signature checks for Retractions of every kind doesn't make much sense. In fact every Retraction is proved only by the signature of its sender (it's different from the Relayable federation, because parent author is allowed to retract comments of any author). And the signature of the sender is checked at the Salmon level. So maybe abolishment of signatures for Retractions (done gradually) won't make any harm. @jhass?

In fact, for the Relayables it's only necessary to pass author signature and parent author signature is redundant for the same reasons - it just repeats Salmon signature check.

cmrd-senya:
I think proposals in my last comment probably need an issue to be opened in the diaspora tracker or here.

cmrd-senya:
For retractions abolishment of signatures can be done pretty softly, by just moving to Retraction.

cmrd-senya:
I've opend an issue on the Diaspora tracker on the signatures problem. It's out of scope here anyway.

cmrd-senya:
@SuperTux88, tell me if it's fine to squash the commits or I have to fix something else.

lib/diaspora_federation/entities/like.rb
```diff
@@ -1,10 +1,30 @@
 module DiasporaFederation
   module Entities
+    # this entity represents a like to some kind of post (e.g. status message)
+    #
+    # @see Validators::LikeValidator
     class Like < Entity
+      # @!attribute [r] positive
+      #   If true set a like, if false, remove it
```
SuperTux88: i thought positive=false is a dislike (which is not implemented, but possible by the protocol)? and removing a like is done by a retraction?

cmrd-senya: True, my mistake. [Here](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/likeable.rb#L9-L10) is what hints to this.

SuperTux88:
I think you can squash it now :)

cmrd-senya:
done

SuperTux88:
thanks :)

PR #18, opened 2015-12-01
Retractions signatures support

cmrd-senya:


lib/diaspora_federation/entities/signed_retraction.rb
```diff
@@ -25,7 +25,48 @@ class SignedRetraction < Entity
...
+        end
+      end
+
+      # verifies the signature +target_author_signature+
+      # @param [Hash] data hash with data to verify
+      # @raise [SignatureVerificationFailed] if the signature is not valid or no public key is found
+      def self.verify_signatures(data)
+        pkey = DiasporaFederation.callbacks.trigger(:fetch_public_key_by_id, data[:diaspora_id])
+        raise SignatureVerificationFailed, "failed to fetch public key for #{data[:diaspora_id]}" if pkey.nil?
+        raise SignatureVerificationFailed, "wrong author_signature" unless Signing.verify_signature(
```
jhass: I'd stop using a modifer-if/unless when you have to break it across several lines. Using the regular variant becomes much cleaner usually at this point.

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -28,6 +28,10 @@ def self.included(model)
           #   a target pod.
           #   @return [String] author signature
           property :author_signature, default: nil
+
+          def self.verify_signatures(data)
```
SuperTux88: what does this?

cmrd-senya: I changed [this](https://github.com/SuperTux88/diaspora_federation/pull/18/files#diff-de8e01cf36c204663baaab85af6af826R102) line to make it support every type of objects that respond to ```verify_signatures ```. Maybe changing Relayable.verify_signatures to class methods of relayable children would be good then to make it respective.

lib/diaspora_federation/entities/relayable_retraction.rb
```diff
@@ -46,7 +46,76 @@ class RelayableRetraction < Entity
...
+        elsif target_author != hash[:diaspora_id] && hash[:parent_author_signature].nil?
+          hash[:parent_author_signature] = Signing.sign_with_key(apply_signable_exceptions(hash), pkey)
+        end
+      end
+      private_class_method :fill_required_signature
+
+      # verifies the signatures (+target_author_signature+ and +parent_author_signature+ if needed)
+      # @param [Hash] data hash with data to verify
+      # @raise [SignatureVerificationFailed] if the signature is not valid or no public key is found
+      def self.verify_signatures(data)
```
SuperTux88: I'm not sure if we should verify this, because we would remove signatures from retractions. I think we should check if the author has the permissions to delete this object and ignore the signatures, then we can remove the signatures here later. But we need to add the signatures for older diaspora pods.

lib/diaspora_federation/entities/relayable_retraction.rb
```diff
@@ -46,7 +46,76 @@ class RelayableRetraction < Entity
...
+            raise SignatureVerificationFailed, "wrong target_author_signature"
+          end
+        elsif !Signing.verify_signature(apply_signable_exceptions(data), data[:parent_author_signature], pkey)
+          raise SignatureVerificationFailed, "wrong parent_author_signature"
+        end
+      end
+      private_class_method :verify_required_signature
+
+      def self.apply_signable_exceptions(data)
+        new_data = data.dup
```
SuperTux88: `data.dup.tap {|data| data.delete(:diaspora_id) }`

lib/diaspora_federation/entities/relayable_retraction.rb
cmrd-senya: Hmm, yes, I think that would simplify the matter. Although the redundancy of the signatures in the retractions is in state of assumption, that still required to be strictly proved, I think we can implement verification in case we find out we really need it in some cases.

> I think we should check if the author has the permissions to delete this object and ignore the signatures

This most probably would be a part of application but not the gem I think.

I'll remove signatures verification then.

SuperTux88: > This most probably would be a part of application but not the gem I think.
I would move this logic to the gem too (when I move the postzord stuff to the gem).

If the sender is the target_author or the parent_author, we need to delete the entity. Even if the retraction wasn't relayed over the parent_author, both are allowed to delete it. So I really think there is no need to verify the signatures.

lib/diaspora_federation/test.rb
```diff
@@ -26,5 +26,23 @@ def self.relayable_attributes_with_signatures(factory_name)
...
+      sort_hash(FactoryGirl.attributes_for(:signed_retraction_entity), Entities::SignedRetraction).tap do |data|
+        Entities::SignedRetraction.update_signatures!(data)
+      end
+    end
+
+    # Generates attributes for relayable retraction entity constructor with correct signatures in it
+    #
+    # @return [Hash] hash with correct signatures
+    def self.relayable_retraction_attributes_with_signatures
+      sort_hash(FactoryGirl.attributes_for(:relayable_retraction_entity), Entities::RelayableRetraction).tap {|data|
```
SuperTux88: please use `do` and `end` here too

lib/diaspora_federation/entities/relayable_retraction.rb
cmrd-senya: Actually, that would mean that any kind of object can be retracted by usual Retraction and SignedRetraction and RelayableRetraction are redundant and have to be deprecated. And if they are deprecated, then we don't want to generate them in the new code, so maybe we don't need signature generating code as well?

I tested retractions in my recent yet unpublished tests. I found out that usual Retraction can be used to delete posts instead of SignedRetraction, but it couldn't be used to delete Relayables in case when the parent object's author tries to retract due to [this](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/federated/retraction.rb#L58) condition. We could change this condition, adding that if ```self.type``` is a relayable, then ```self.target.author``` is allowed to be an author of the parent of the target. Then Retraction could replace all the other types of retractions. And leave RelayableRetraction and SignedRetraction only for the case of receive from legacy pods and fetch them without checking inner signatures.

cmrd-senya: Well, we still need RelayableRetraction for older pods due to the thing I mentioned... And we have to study how older pods work before we can change generation of the objects. So, I think we have to leave signatures update for now.

cmrd-senya: Looks like I'm talking to myself, oh

SuperTux88: LOL :D

But yes:

* We should make them depricated.
* We can use `Retraction` instead of `SignedRetraction` if the current diaspora can handle this. But we should also test if this is compatible with friendica (cc @annando).
* We still need `SignedRetraction` and `RelayableRetraction` to receive them from old diaspora pods.

annando: Am 02.12.15 um 02:21 schrieb Benjamin Neff:

>   * We can use |Retraction| instead of |SignedRetraction| if the current
>     diaspora can handle this. But we should also test if this is
>     compatible with friendica (cc @annando <https://github.com/annando>).

Friendica can receive retractions, signed_retractions and
relayable_retractions:
https://github.com/friendica/friendica/blob/master/include/diaspora.php#L95-L103

I haven't worked with retractions by now. Just tell me what they are
expected to do.

By now it seems as if received retractions are supported for persons and
posts. We are sending them from Friendica only for persons. for posts we
use the signed ones.

Michael

Flaburgan: I would be more confident about dropping SignedRetraction if @maxwellsalzberg could confirm us why it was introduced at the first time.

cmrd-senya: Then it's better to point him there for the discussion:
https://github.com/diaspora/diaspora/issues/6566

SuperTux88: @annando: Diaspora currently sends `signed_retraction`s for Posts and `relayable_retraction` for Relayables and `retraction` for everything else (Photo).

When do you send a retraction for a person? For persons there is a `account_deletion` message.

What we want to change as first step (hopefully by the next diaspora release in 6 weeks) is, that we send `retraction`s instead of `signed_retraction` for posts. Diaspora is able to parse this correctly. The question is: Can friendica handle a `retraction` for a post? Or should we wait with this change?

@Flaburgan: there are currently many things in the federation code, that aren't needed, but were added in the past. But you can't remove them simply, because you need to stay compatible (thats why they are still there, nobody touched these things since years). I want to clean the federation, to make it less complex, more stable and easier to implement by others (friendica/redmatrix). But we need many iterations for this, because I don't want to break something.

cmrd-senya: @SuperTux88, yes, there is one more case when Retraction is sent for a user (person), looks like I missed it before. [Here](https://github.com/diaspora/diaspora/blob/develop/app/models/user/connecting.rb#L57) it is.

We have to make sure, that every non-abandonded pod in the federation would support receiving Retraction instead of SignedRetraction in order to do that.

Also, I already wrote a patch for Diaspora and will publish it as soon as this is merged, that allows to Retract relayables with usual Retraction. After all the federation will move on the patched version we can stop sending RelayableRetraction as well.

SuperTux88: OK, retraction for a person is when you stop sharing with someone...

cmrd-senya: We'll have to document it as well.

lib/diaspora_federation.rb
```diff
@@ -22,10 +22,11 @@ module DiasporaFederation
     fetch_person_for_hcard
     save_person_after_webfinger
     fetch_private_key_by_id
-    fetch_private_key_by_post_guid
+    fetch_private_key_by_entity_guid
     fetch_public_key_by_id
-    fetch_public_key_by_post_guid
-    post_author_is_local?
+    fetch_public_key_by_entity_guid
```
SuperTux88: I think we should rename this to `fetch_author_public_key_by_entity_guid` to make it clearer (because its the author of the entity). And I also think that we should rename `fetch_public_key_by_id` to `fetch_public_key_by_diaspora_id` because id is to generic. (and the same for private key) What do you think?

and regardless of how good we name this callbacks, we should also document them somewhere (module-documentation of `DiasporaFederation`?). What the callbacks should do and what parameters they have.

cmrd-senya: I don't mind renaming it.

I think, the other place to document the callbacks is README.md.

lib/diaspora_federation/entities/relayable_retraction.rb
```diff
@@ -46,7 +46,45 @@ class RelayableRetraction < Entity
...
-      property :target_author_signature
+      property :target_author_signature, default: nil
+
+      # Generates XML and updates signatures
+      # @see Entity#to_xml
+      # @return [Nokogiri::XML::Element] root element containing properties as child elements
+      def to_xml
+        entity_xml.tap do |xml|
+          hash = to_h
+          self.class.update_signatures!(hash)
```
SuperTux88: Why not `RelayableRetraction.update_signatures!(hash)` ?

lib/diaspora_federation.rb
SuperTux88: I also thought about the README.md, but I like the `DiasporaFederation` module more. I don't want the readme too big, because it should tell some basic info about the gem/project and the technical documentation is in the [rubydoc](http://www.rubydoc.info/gems/diaspora_federation/) then.

cmrd-senya: I think ```DiasporaFederation``` module is fine.

lib/diaspora_federation.rb
```diff
@@ -22,10 +22,11 @@ module DiasporaFederation
...
+    entity_author_is_local?
+    fetch_entity_author_id_by_guid
```
cmrd-senya: Do you think we need renaming this to ```fetch_entity_author_diaspora_id_by_guid```? I think it is clear here as is.

lib/diaspora_federation/entities/relayable_retraction.rb
cmrd-senya: Does it make any difference?

lib/diaspora_federation.rb
SuperTux88: no, I think author_id is good.

lib/diaspora_federation/entities/relayable_retraction.rb
SuperTux88: I was confused why it is dynamic.

SuperTux88: in `Relayable` it is also `Relayable.update_signatures!(hash)`. I think if `self.class` always is the same, we should use the class name directly.

lib/diaspora_federation/entities/relayable_retraction.rb
Flaburgan: >there are currently many things in the federation code, that aren't needed, but were added in the past. But you can't remove them simply, because you need to stay compatible (thats why they are still there, nobody touched these things since years). I want to clean the federation, to make it less complex, more stable and easier to implement by others (friendica/redmatrix). But we need many iterations for this, because I don't want to break something.

Yeah sure! My point was, how are we sure that they really *"aren't needed"*, if @maxwell can confirm I would be reassured :) Moreover, there maybe are things in the federation put there planning to use it for future features. So because it's hard to keep every pod out there able to communicate, I want to be sure we're not dropping something which would have been useful later ;)

annando: Am 02.12.15 um 17:45 schrieb Benjamin Neff:

> When do you send a retraction for a person? For persons there is a
> |account_deletion| message.

We send this when a Friendica user is unfollowing a Diaspora contact.
The XML looks like this:

<XML>
  <post>
    <retraction>
      <post_guid>{{$guid}}</post_guid>
      <diaspora_handle>{{$handle}}</diaspora_handle>
      <type>Person</type>
    </retraction>
  </post>
</XML>

{{$guid}} = The GUID of the Friendica user
{{$handle}} = The Address of the Friendica user

> What we want to change as first step (hopefully by the next diaspora
> release in 6 weeks) is, that we send |retraction|s instead of
> |signed_retraction| for posts. Diaspora is able to parse this correctly.
> The question is: Can friendica handle a |retraction| for a post? Or
> should we wait with this change?

It looks as if there is code - but this code doesn't look like it could
work. Could you provide me an XML with a post retraction? Then I can
compare it. The good thing is: We are currently preparing our new
release and will do a feature freeze in these days. This will be part of
the release that is planned for Christmas.

BTW: Are you attending the 32C3? If yes then I would like to grep you
for an hour or so. Then we could make a review of the Diaspora
implementation in Friendica.

cmrd-senya: This can be used to retract a post
```
<XML>
  <post>
    <retraction>
      <post_guid>8f142f807b3901333e9c0800276d7a8d</post_guid>
      <diaspora_handle>alice@localhost:9887</diaspora_handle>
      <type>StatusMessage</type>
    </retraction>
  </post>
</XML>
```

SignedRetraction is currently used for retracting posts but I don't have XML on hand ATM.

cmrd-senya:
I fixed the discussed issues, except for documentation. I think it could be added in a different PR, it is out of scope for that one.

lib/diaspora_federation/entities/relayable_retraction.rb
SuperTux88: @annando: yes I'm at the 32C3.

SuperTux88:
Yes, callback-documentation can be added later. Thanks.

lib/diaspora_federation/entities/relayable_retraction.rb
annando: Am 02.12.15 um 22:27 schrieb Senya:

> SignedRetraction is currently used for retracting posts but I don't have
> XML on hand ATM.

I will add support for receiving it in this format. Question is what we
should send from Friendica. Do (most) Diaspora servers already
understand the unsigned retractions?

cmrd-senya: We will answer this shortly, follow
https://github.com/diaspora/diaspora/issues/6566

maxwell: Sorry I'm exactly zero help here.

On Wed, Dec 2, 2015 at 1:03 PM, Fla <notifications@github.com> wrote:

> In lib/diaspora_federation/entities/relayable_retraction.rb
> <https://github.com/SuperTux88/diaspora_federation/pull/18#discussion_r46474759>
> :
>
> > +        fill_required_signature(target_author, pkey, hash) unless pkey.nil?
> > +      end
> > +
> > +      def self.fill_required_signature(target_author, pkey, hash)
> > +        if target_author == hash[:diaspora_id] && hash[:target_author_signature].nil?
> > +          hash[:target_author_signature] = Signing.sign_with_key(apply_signable_exceptions(hash), pkey)
> > +        elsif target_author != hash[:diaspora_id] && hash[:parent_author_signature].nil?
> > +          hash[:parent_author_signature] = Signing.sign_with_key(apply_signable_exceptions(hash), pkey)
> > +        end
> > +      end
> > +      private_class_method :fill_required_signature
> > +
> > +      # verifies the signatures (+target_author_signature+ and +parent_author_signature+ if needed)
> > +      # @param [Hash] data hash with data to verify
> > +      # @raise [SignatureVerificationFailed] if the signature is not valid or no public key is found
> > +      def self.verify_signatures(data)
>
> there are currently many things in the federation code, that aren't
> needed, but were added in the past. But you can't remove them simply,
> because you need to stay compatible (thats why they are still there, nobody
> touched these things since years). I want to clean the federation, to make
> it less complex, more stable and easier to implement by others
> (friendica/redmatrix). But we need many iterations for this, because I
> don't want to break something.
>
> Yeah sure! My point was, how are we sure that they really *"aren't
> needed"*, if @maxwell <https://github.com/maxwell> can confirm I would be
> reassured :) Moreover, there maybe are things in the federation put there
> planning to use it for future features. So because it's hard to keep every
> pod out there able to communicate, I want to be sure we're not dropping
> something which would have been useful later ;)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SuperTux88/diaspora_federation/pull/18/files#r46474759>
> .
>



-- 
maxwell salzberg - BackerKit, founder
skype: max.salzberg
p: (203) 451-3600
email: maxwell@backerkit.com
twitter: @maxwellsalz

PR #20, opened 2015-12-13
Receiving entities support

cmrd-senya:
Introduces DiasporaFederation::Postzord::Receive.

Probably lacks some tests.

### DIFF
--- a/lib/diaspora_federation/receiver/private.rb
+++ b/lib/diaspora_federation/receiver/private.rb
@@ -1,0 +1,24 @@
+module DiasporaFederation
+  # RecipientNotFound is raised when failed to fetch a private key of the recipient of the received message
+  class RecipientNotFound < Exception
+  end
+
+  class Receiver
+    # Receiver::Private is used to receive private messages, which are addressed to a specific user, encrypted with his
+    # public key and packed using Salmon::EncryptedSlap
+    class Private < Receiver
+      def initialize(recipient_guid, salmon_xml)
+        super(salmon_xml)
+        @recipient_guid = recipient_guid
+        @recipient_private_key = DiasporaFederation.callbacks.trigger(:fetch_private_key_by_user_guid, recipient_guid)
+        raise RecipientNotFound if @recipient_private_key.nil?
+      end
+
+      protected
+
+      def slap
+        @salmon ||= Salmon::EncryptedSlap.from_xml(@salmon_xml, @recipient_private_key)
+      end
+    end
+  end
+end
--- a/spec/lib/diaspora_federation/receiver/private_spec.rb
+++ b/spec/lib/diaspora_federation/receiver/private_spec.rb
@@ -1,0 +1,57 @@
+module DiasporaFederation
+  describe Receiver::Private do
+    let(:sender_id) { FactoryGirl.generate(:diaspora_id) }
+    let(:sender_key) { OpenSSL::PKey::RSA.generate(1024) }
+    let(:recipient_key) { OpenSSL::PKey::RSA.generate(1024) }
+    let(:recipient_guid) { FactoryGirl.generate(:guid) }
+    let(:xml) {
+      DiasporaFederation::Salmon::EncryptedSlap.generate_xml(
+        sender_id,
+        sender_key,
+        FactoryGirl.build(:request_entity),
+        recipient_key
+      )
+    }
+
+    it "calls entity_persist if everyting is fine" do
+      expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                                .with(:fetch_public_key_by_diaspora_id, sender_id)
+                                                .and_return(sender_key)
+      expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                                .with(:fetch_private_key_by_user_guid, recipient_guid)
+                                                .and_return(recipient_key)
+
+      expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                                .with(:entity_persist, kind_of(Entity), recipient_guid, sender_id)
+
+      described_class.new(recipient_guid, xml).receive!
+    end
+
+    it "raises when sender public key is not available" do
+      expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                                .with(:fetch_public_key_by_diaspora_id, sender_id)
+                                                .and_return(nil)
+      expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                                .with(:fetch_private_key_by_user_guid, recipient_guid)
+                                                .and_return(recipient_key)
+
+      expect { described_class.new(recipient_guid, xml).receive! }.to raise_error SenderNotFound
+    end
+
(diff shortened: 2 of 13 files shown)


diaspora_federation.gemspec
```diff
@@ -26,4 +26,5 @@ Gem::Specification.new do |s|
   s.add_dependency "faraday_middleware", "~> 0.10.0"
   s.add_dependency "typhoeus", "~> 0.7"
   s.add_dependency "valid", "~> 1.0"
+  s.add_dependency "json"
```
jhass: `json` is in the standard library.

SuperTux88:
I think we can remove the "Postzord", because I don't really know what this word means and it is not only for "Posts" ...

Simply move the receiver one level up.

cmrd-senya:
Done that.

I added tests and docs also.

lib/diaspora_federation/receiver/private.rb
```diff
@@ -0,0 +1,24 @@
+module DiasporaFederation
+  # RecipientNotFound is raised when failed to fetch a private key of the recipient of the received message
+  class RecipientNotFound < Exception
+  end
+
+  class Receiver
+    # Receiver::Private is used to recieve private messages, which are addressed to a specific user, encrypted with his
```
SuperTux88: "... receive ..."

lib/diaspora_federation/receiver/private.rb
```diff
@@ -0,0 +1,24 @@
...
+    # public key and packed using Salmon::EncryptedSlap
+    class Private < self
```
SuperTux88: why `< self` ? what does this?

spec/lib/diaspora_federation/receiver/private_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        sender_key,
+        FactoryGirl.build(:request_entity),
+        recipient_key
+      )
+    }
+
+    it "calls entity_persist if everyting is fine" do
+      expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                                .with(:fetch_public_key_by_diaspora_id, sender_id)
+                                                .once
```
SuperTux88: I think `once` is default.

lib/diaspora_federation/receiver/private.rb
cmrd-senya: It inherits Private from Receiver, does it?

spec/lib/diaspora_federation/receiver/private_spec.rb
cmrd-senya: Then why do they need the specifier? I though default is any times.
There is no specifier for that.

lib/diaspora_federation/receiver/private.rb
SuperTux88: thats what I thought ;) but why not `< Receiver`? I think that's better.

spec/lib/diaspora_federation/receiver/private_spec.rb
SuperTux88: I don't know, but if I remove the `once` and do something twice it fails with `expected: 1 time` and `received: 2 times`.

cmrd-senya: Ok, then :)

cmrd-senya:
Changes are applied.
However I still want to know why do they have ```once``` specifier if it is a default :)

cmrd-senya:
TBH, I'm not sure the receive controller implemented in this PR is what Diaspora really needs, because Diaspora launches a worker when receives an xml and only the worker launches real xml parsing and verification. I guess this was made so that the receive controller ends its work cycle for every xml ASAP and pass it to a queue.  So we would propably want to allow such behaviour.

cmrd-senya:
Ok.

I had a little chat with @jhass, where he explained me that we need workers on receive process because we must minimize the time (which is around 2 seconds per message) we hold a sender because we have a limitation on unicorn workers. Now in the Diaspora code all the operation after receiving HTTP request is performed in a worker process, including XML parsing of the federation message.

I did a little profiling test with [this](https://gist.github.com/cmrd-senya/3accb933c5a6a78979f6) patch. The result (I found no place to upload an 15 Mb html page yet) shows that we spend no more than 22% of time on xml parsing, signature verification, etc. Application logic, like notifications, DB store takes around 78%. 15.79% time is call to Diaspora::Parser#from_xml. 13.39% is ROXML::ClassMethods::Operations#from_xml. And ROXML would be dropped as soon as we merge DiasporaFederation module, so probably the new parsing code would run faster.

I think, that therefore we can move xml parsing to the unicorn process scope. This would allow:
- More proper status codes if a message is ill-composed. For example, we could return an error code if a signature is wrong and not 200 as now.
- The flow logic simplifies. Thus, we would need only one callback ```:entity_persist``` on receive, which would pass a job to a worker in Diaspora. Otherwise we would require two callbacks, one to call after receiving XML, and another after parsing.

This way is how it is implemented in this PR. I think, it is likely that overheads could be reduced as we merge DiasporaFederation, so the grow is not very significant.

SuperTux88:
Sorry, I didn't have much time the last days (to much work ...).

I'll merge this now, but I think we should keep the processing out of the unicorn-workers. Because if you make the parsing in the request-scope, the persist will also be done in the request scope, and if this takes 78% of the time, then we should do this in the backend. Also we can do retries for failures if it is a backend job.

SuperTux88:
I added the callbacks for this in the `develop` branch and did a few refactorings.

cmrd-senya:
Why can't we do XML parsing in the receive scope and persisting in the
queue? That would simplify the library and make error processing better.

SuperTux88:
you can not only do the persist in the queue, because you need to convert the entity again to a string, to send it to the queue and then parse it again in diaspora and then persist it. It is simpler to do the whole parse process in the queue and then you can persist it directly.

cmrd-senya:
Why do I need to convert it to a string? I can form an instance of
Entities::Something and pass it to callbacks.trigger(:entity_persist,
something_instance). This way it work in the PR.

SuperTux88:
yes, but if you want to process the entity in the queue after pass it to `callbacks.trigger(:entity_persist, something_instance)` you need to transform it to a string to send it to the queue and parse it again in the job. It's simpler to pass the xml-string to the queue and do the whole parsing in the background job

cmrd-senya:
Isn't there a way to pass an object to the queue worker scope?

SuperTux88:
No, because these are different processes. The parameter for the jobs are passed over a redis database.

cmrd-senya:
That's a pity. However, redis support hashes, and converting to and from
hash makes more sense in our case and not that big deal, but anyway,
it's worse that object passing.

I really dislike returning 200 on some clearly faulty objects.

denschub:
I really think parsing in the worker makes more sense since it could possible take a while, consume a lot of CPU and workers are easier to scale.

PR #21, opened 2015-12-21
Allow parent entity types different than "Post"

cmrd-senya:
Entities::Message requires "Conversation" at the place, for instance.


SuperTux88:
thanks, good catch

PR #22, opened 2015-12-23
HydraWrapper fetched from the Diaspora source

cmrd-senya:
I think it's more of [WIP] thing and we are more concentrated on the receive now, but I think it's no harm to make a PR now, let it wait for its time.

### DIFF
--- a/lib/diaspora_federation.rb
+++ b/lib/diaspora_federation.rb
@@ -186,3 +186,5 @@ module DiasporaFederation
   class ConfigurationError < RuntimeError
   end
 end
+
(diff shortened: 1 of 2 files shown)


Flaburgan:
@DenSchub recently closed https://github.com/diaspora/diaspora/pull/5217 saying to should be done in this gem instead of inside the diaspora* code, is this something which should be integrated in this PR?

cmrd-senya:
I beleive this PR is outdated now, since @SuperTux88 has integrated HydraWrapper.

Maybe then create a new issue for diaspora_federation?

Issue #15, opened 2015-11-22
Making a 0.0.9 release


cmrd-senya:
Looks like the functionality is now enough for me to finish tests at [6539](https://github.com/diaspora/diaspora/pull/6539). I want it to be merged to Diaspora and this will definitely require to rely on a release version of the diaspora_federation gem. So, can we form TODOs required for a new (though minor) release?

SuperTux88:
The main-TODO is to add documentation to all new entities and validators.

Currently this looks like this:
```
Modules:        14 (    3 undocumented)
Classes:        84 (   35 undocumented)
Constants:      22 (    0 undocumented)
Methods:       139 (    6 undocumented)
 83.01% documented
```

You can see this when you run `yardoc --list-undoc`.

cmrd-senya:
So can we release 0.0.9 now?
@jhass, since it is a blocker for the [6539](https://github.com/diaspora/diaspora/pull/6539) I think it would be nice if you confirm it is ok and won't require any changes in the gem to merge the PR.

SuperTux88:
I want to have a look at the gem and maybe do some cleanups today or tomorrow, but then I can build a release.

cmrd-senya:
I think you'll merge salmon to master right? salmon was our development branch, so we'll need a new one then with a more straight name.

SuperTux88:
yes, I'll create a `develop` branch after that. `salmon` was my feature-branch as I was working alone at this ;)

cmrd-senya:
@SuperTux88, I just thought that the next thing I would need are retractions and they have to support signatures as well in the gem. I think that I want to see them in the release too. So how about setting this as a blocker for the release?

SuperTux88:
Add whatever you need and then say me when you need a release :)

SuperTux88:
just for your information: I moved the `salmon` branch to `develop` and made it the default branch.

cmrd-senya:
I think time for the release has come

SuperTux88:
I'll do the release at the weekend.

SuperTux88:
I just released 0.0.9: https://rubygems.org/gems/diaspora_federation

cmrd-senya:
Thanks!

Flaburgan:
Awesome guys :)

I'm wondering, you are using semver here, why do you increment the hotfix number instead of the minor one?

cmrd-senya:
According to semver numeration must start with 0.1.0, so we have to
[...]

PR #24, opened 2015-12-30
fix created_at in factories

cmrd-senya:


SuperTux88:
Please don't create release-PRs, I'll create releases when needed. Also I don't want to cherry-pick commits for the releases, I always release the whole develop branch (except if it is something security-relevant). Because of that I didn't look at #23 in detail, and thought everything you need will be released if I release the develop branch. I didn't see that the PR contains changes that aren't in develop yet. So please create your changes as normal PRs.

Also: diaspora uses UTC in the database and also in federation. Where does it need local timezone?

cmrd-senya:
You must have explained your point before. That's why I wanted a discussion. I was unsure about what you had on your mind. Just silently ignoring something you don't agree with is unproductive and impolite as long as I post requests for reason. I can't post changes without them being reviewed, but I didn't have a chance to review your changes when you made it. Otherwise I would notice something important lacks from the release.

I can drop the release commit, but I need a release to merge https://github.com/diaspora/diaspora/pull/6595. I also was going to propose to release a version 0.1.0 when we would have merge receive updates, because it makes more sense for semver and would mark some sort of important change.

It's not about timezones, it is about string format returned by different functions. Probably we don't need ```Time.zone.now``` then, but something else, but ```DateTime.now.utc``` returns a string in a format that doesn't match with the format produced by Diaspora

cmrd-senya:
That's what I talk about

SuperTux88:
I already said, that I wasn't really active in IRC the last few days, sorry for that. And sorry that I didn't see that #23 contained additional commits, I've only read the title and thought it is something like #15 but only with a few commits. So I thought if I release all commits from develop, everything is fine.

You're welcome to give feedback to my commits, if you want to review them.

I will release 0.1.0 if the gem is fully functional integrated in diaspora.

If it is only the string-format, maybe you can use `Time.now.utc` like in the entities as default?

cmrd-senya:
Looks like it is ok with the default.

cmrd-senya:
I think that fully functional gem integrated to Diaspora must be called 1.0.0 version according to semver. Semver says that 0.0.* versioning is not cool.

SuperTux88:
using the default breaks the tests. but you can use the same as for default

semver doesn't make much sense at the current time of development. there are changes that need changes in diaspora in every release.

1.0.0 will be the version when all todos and cleanups are done.

cmrd-senya:
Updated

SuperTux88:
merged, thx. I'll release this today evening when I'm at home.

cmrd-senya:
Great! Thanks for that!

PR #42, opened 2016-12-04
Refactor Entity API: access signatures for relayables

cmrd-senya:
Please, don't consider changes in this PR as clear, completed and mergable. I'm posting them in order to illustrate possible solutions to the issues I described here.

I'm currently working on user archive export improvements, and particularly on including other people's data (comments, etc) to the archive (as was decided before on Loomio). In order to export a comment of another person, we have to have a signature for that comment. If both comment and parent authors are local, then the comment signature doesn't get stored on a pod and instead gets generated every time on the send. Since it is now a responsibility of the diaspora_federation gem to generate signatures, in order to get a signature I have to convert a comment to a comment federation entity with this spell first: `Diaspora::Federation::Entities.build(object)`. It makes perfect sense: archive backup can be considered as a kind of federation, but "offline".

What I finally want to achieve, is to be able to get a JSON object for a comment (or like, etc) with a signature in it for the export archive, so a remote pod which performs a back up from that archive can trust this comment. So it differs from the way it works now by a format of an output data mostly. I need JSON instead of XML. I also don't need legacy XML name substitution.

Currently, API that is provided by `DiasporaFederation::Entity` is a bit confusing. The only public method which I can use to get signatures is `#to_xml`. But it gives me data in XML format which isn't what I want. There is also a `#to_h` method which gives me a hash, but it doesn't run signature generation flow for relayables. `#to_h` is actually does recursive call to to_h to all non-string objects of an entity.

There is also a private method called `#xml_elements`. Its name is prefixed with `xml_` apparently because it is used in the XML generation routine. However, the method itself has nothing to do with XML. It takes properties of the entity and converts them to strings where the type is appropriate. The value here for me is that it is the method that gets overriden for relayable entities to return signatures besides. Basically, this method solves my issue, because calling `.to_json` on it's output gives me a desired result.

In general, `#to_h` is not very useful as a public API method. I expect that public API returns me complete data, and `#to_h` doesn't do it for every case (like with signatures). So my first proposal is to rename it to something less promising, for instance `#unfold`, and make it protected, because you can't make much use of it outside the internals of diaspora_federation.

The second thing I propose is to rename `#xml_elements` to something unrelated to XML, for instance `#normalized_properties`. It better represents what it does: patches properties to something that fits better to final rendering.

The third thing I propose is to introduce a method `DiasporaFederation::Entity#to_json`. It simply renders `#normalized_properties` to JSON, besides a little addition: it adds a `entity_class` property to the returned JSON object that plays the same role as the root element of an XML entity-object does: it helps to distinguish one entity type from another. There is no such thing as a "name" for a JSON object, so I guess it's fine to store it among other properties. Having the object type as "root" like `"comment": {}` is also not an option, since the `"comment"` part must be a property of some higher parent object which semantically doesn't make sense.

I believe these three modifications will help to make the Entity API a little more intelligible and also solve my issue :) I'm ready to complete this PR once we are generally agree on the changes we are fine with.

### DIFF
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -69,23 +69,6 @@ module DiasporaFederation
       validate
     end
 
-    # Returns a Hash representing this Entity (attributes => values).
-    # Nested entities are also converted to a Hash.
-    # @return [Hash] entity data (mostly equal to the hash used for initialization).
-    def to_h
-      properties.map {|key, value|
-        type = self.class.class_props[key]
-
-        if type == String || value.nil?
-          [key, value]
-        elsif type.instance_of?(Class)
-          [key, value.to_h]
-        elsif type.instance_of?(Array)
-          [key, value.map(&:to_h)]
-        end
-      }.to_h
-    end
-
     # Returns the XML representation for this entity constructed out of
     # {http://www.rubydoc.info/gems/nokogiri/Nokogiri/XML/Element Nokogiri::XML::Element}s
     #
@@ -147,6 +130,25 @@ module DiasporaFederation
     # @return [String] string representation of this object
     def to_s
       "#{self.class.name.rpartition('::').last}#{":#{guid}" if respond_to?(:guid)}"
+    end
+
+    protected
+
+    # Returns a Hash representing this Entity (attributes => values).
+    # Nested entities are also converted to a Hash.
+    # @return [Hash] entity data (mostly equal to the hash used for initialization).
+    def unfold
+      properties.map {|key, value|
+        type = self.class.class_props[key]
+
+        if type == String || value.nil?
+          [key, value]
+        elsif type.instance_of?(Class)
+          [key, value.unfold]
+        elsif type.instance_of?(Array)
+          [key, value.map(&:unfold)]
+        end
+      }.to_h
     end
 
     private
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -79,7 +79,7 @@ module DiasporaFederation
     def to_xml
       doc = Nokogiri::XML::DocumentFragment.new(Nokogiri::XML::Document.new)
       Nokogiri::XML::Element.new(self.class.entity_name, doc).tap do |root_element|
-        xml_elements.each do |name, value|
+        normalized_properties.each do |name, value|
           add_property_to_xml(doc, root_element, name, value)
         end
       end
@@ -217,7 +217,7 @@ module DiasporaFederation
       end
     end
 
-    def xml_elements
+    def normalized_properties
       properties.map {|name, value| [name, self.class.class_props[name] == String ? value.to_s : value] }.to_h
     end
 
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -85,6 +85,10 @@ module DiasporaFederation
       end
     end
 
+    def to_json
+      normalized_properties.merge!(entity_class: self.class.entity_name).to_json
+    end
+
     # Construct a new instance of the given Entity and populate the properties
     # with the attributes found in the XML.
     # Works recursively on nested Entities and Arrays thereof.
--- a/lib/diaspora_federation/entities/relayable.rb
+++ b/lib/diaspora_federation/entities/relayable.rb
@@ -151,7 +151,7 @@ module DiasporaFederation
       # if the signatures are not there yet and if the keys are available.
       #
       # @return [Hash] sorted xml elements with updated signatures
-      def xml_elements
+      def normalized_properties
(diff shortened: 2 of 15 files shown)


SuperTux88:
Sorry, didn't have the time to give feedback earlier. So, where should I begin?

First: The `#to_h` method was meant to be public (for example for something like json), so I would keep it. But we should change it, so it includes the generated signatures. (that means more or less `#xml_elements` -> `#to_h` and make it public)

Second: `#normalized_properties` makes sense, but only in the entity.rb, that can then be used in `#signature_data` in relayable.rb as replacement for `#to_h` (if `#to_h` adds signatures, it can't be used as source for signature data ;) ).

And last: I'm unsure about the `#to_json` part. The federation is XML, so I don't know if we should support that from the federation code, or if we should do that part in diaspora. But why not (I don't have a good reason against it ;) ). But if we add JSON support, we need also a `#from_xml`, and we need to document the JSON format in the federation documentation.

About the JSON format: I would name the type-key `entity_name` (like the method) or maybe `entity_type`, but not `entity_class`. Also you need to add the order for relayables, because JSON is unordered.

lib/diaspora_federation/entity.rb
```diff
@@ -96,12 +79,16 @@ def to_h
     def to_xml
       doc = Nokogiri::XML::DocumentFragment.new(Nokogiri::XML::Document.new)
       Nokogiri::XML::Element.new(self.class.entity_name, doc).tap do |root_element|
-        xml_elements.each do |name, value|
+        normalized_properties.each do |name, value|
```
SuperTux88: Use `to_h` here. And `#to_h` should use `normalized_properties` instead of `properties`.

lib/diaspora_federation/entity.rb
```diff
@@ -96,12 +79,16 @@ def to_h
...
           add_property_to_xml(doc, root_element, name, value)
         end
       end
     end
 
+    def to_json
+      normalized_properties.merge!(entity_class: self.class.entity_name).to_json
```
SuperTux88: Use `to_h` here.

lib/diaspora_federation/entity.rb
```diff
@@ -215,7 +221,7 @@ def properties
       end
     end
 
-    def xml_elements
+    def normalized_properties
```
SuperTux88: keep that 👍

cmrd-senya:
Now when #43 has been merged, diaspora_federation doesn't internally depend on the `to_h` method anymore. But still, if I create, for instance, a Comment entity without signatures, and call the `to_h` method, it'll return a hash with 
```
 :author_signature=>nil,
 :parent_author_signature=>nil,
```

Do we need this method at all? Maybe it is better to alias `to_h` to `enriched_properties`? I imagine it is more useful to publish methods that return full data, rather than some part of it. In that case, it is possible to implement a `to_json` method at the client side, basing on `to_h`. We still don't have a public method that gives us a hash with signatures. One may want it to implement some unusual entity serialization format (e.g. some binary format, like google protobuf) without a requirement to modify diaspora_federation source.

I see `to_h` is largerly used in the rspec tests, but I can rewrite, perhaps. I didn't take a close look yet, but I'm ready to work on it.

What do you think, @SuperTux88 ?

SuperTux88:
> Maybe it is better to alias `to_h` to `enriched_properties`?

`to_h` currently returns a recursive hash, while `enriched_properties` still contains nested entities, so we can't alias it. But we can use `enriched_properties` in `to_h` instead of `properties`.

> In that case, it is possible to implement a `to_json` method at the client side, basing on to_h.

As already written, I'm unsure if we should add JSON support to the federation library or if we should do it inside of diaspora. But if we base the migration archive on the federation-entities in JSON format it is probably a good idea to add `to_json`/`from_json` to it and also add the JSON-serialization to the documentation (and mention there that it is only used for the migration, not for federation). But anyway, if you have a good reason for what to do I'm open for it :)

cmrd-senya:
> to_h currently returns a recursive hash, while enriched_properties still contains nested entities, so we can't alias it. But we can use enriched_properties in to_h instead of properties.

Okay, I can do that.

> As already written, I'm unsure if we should add JSON support to the federation library or if we should do it inside of diaspora. But if we base the migration archive on the federation-entities in JSON format it is probably a good idea to add to_json/from_json to it and also add the JSON-serialization to the documentation (and mention there that it is only used for the migration, not for federation). But anyway, if you have a good reason for what to do I'm open for it :)

I think when I have https://github.com/diaspora/diaspora/pull/6726 stabilized, it is possible to port `to/from_json` methods here. The logic is not as straighforward as just doing `.enriched_properties.to_json`. E.g. for relayables I'm trying to achieve the order awareness by using [the array of objects](http://stackoverflow.com/a/4515863) approach.

cmrd-senya:
Superseded by #43 and #48.

PR #43, opened 2016-12-28
Add property types and change timestamp format

SuperTux88:
Make properties of the entities aware of different types (String, Integer, Boolean and Time), so they can be generated and parsed in the correct format. This changes the format of timestamps to ISO 8601 (except for `Message` for backward-compatibility, because old Pods need the old format for valid signatures, this can be changes when we drop compatibility with pods older than 0.6).

We need this change to have the new timestamp format for diaspora/diaspora#4269, so that signatures with the timestamp are valid with the new timestamp format.

This PR also contains some changes from #42, `#enriched_properties` can be used to generate JSON with signatures.

### DIFF
--- a/lib/diaspora_federation/discovery/h_card.rb
+++ b/lib/diaspora_federation/discovery/h_card.rb
@@ -43,16 +43,16 @@ module DiasporaFederation
       # @!attribute [r] guid
       #   @see Entities::Person#guid
       #   @return [String] guid
-      property :guid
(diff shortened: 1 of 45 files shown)


cmrd-senya:
I didn't check the source yet. Can you tell if it deprecates #42?

SuperTux88:
It deprecates the first two comments, so you can change #42 to add the to/from JSON functionality.

PR #46, opened 2017-01-04
Fix boolean pattern parse

cmrd-senya:
Fixes the regular expression for booleans

### DIFF
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -319,8 +319,8 @@ module DiasporaFederation
       when :integer
         text.to_i if text =~ /^\d+$/
       when :boolean
-        return true if text =~ /(true|t|yes|y|1)$/i
-        false if text =~ /(false|f|no|n|0)$/i
+        return true if text =~ /^(true|t|yes|y|1)$/i
+        false if text =~ /^(false|f|no|n|0)$/i
       else
         text
       end
--- a/spec/lib/diaspora_federation/entity_spec.rb
+++ b/spec/lib/diaspora_federation/entity_spec.rb
@@ -195,16 +195,28 @@ XML
 
         it "parses boolean fields with false value" do
           xml = <<-XML.strip
-<test_default_entity>
-  <test1>qwer</test1>
-  <test2>qwer</test2>
-  <test3>false</test3>
-</test_default_entity>
+<test_entity_with_boolean>
+  <test>false</test>
+</test_entity_with_boolean>
 XML
 
-          entity = Entities::TestDefaultEntity.from_xml(Nokogiri::XML::Document.parse(xml).root)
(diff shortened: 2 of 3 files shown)


spec/lib/diaspora_federation/entity_spec.rb
```diff
@@ -195,16 +195,28 @@ module DiasporaFederation
...
+          expect(entity.test).to eq(false)
+        end
+
+        it "parses boolean fields with a randomly matching pattern as erroneous" do
+          %w(ttFFFtt yesFFDSFSDy noDFDSFFDFn fXf LLyes).each do |weird_value|
+            xml = <<-XML.strip
+<test_entity_with_boolean>
+  <test>#{weird_value}</test>
+</test_entity_with_boolean>
+            XML
```
SuperTux88: Oh, one little thing, can you remove the spaces in this line?

cmrd-senya: Done.
It must have been my IDE.

SuperTux88: Yes, it was the IDE, I know that ;) Thanks :)

SuperTux88:
Thanks ❤️

lib/diaspora_federation/entity.rb
```diff
@@ -319,8 +319,8 @@ def simple_node(doc, name, value)
       when :integer
         text.to_i if text =~ /^\d+$/
       when :boolean
-        return true if text =~ /(true|t|yes|y|1)$/i
-        false if text =~ /(false|f|no|n|0)$/i
+        return true if text =~ /^(true|t|yes|y|1)$/i
+        false if text =~ /^(false|f|no|n|0)$/i
```
jhass: Careful here, in Ruby `^`and `$`are beginning/end of **line** respectively, regardless of `/m`. If Beginning/end of **string** is intended here, use `\A`and `\z`respectively.

cmrd-senya: Is it a real issue here? We can declare multiline value for booleans, integers and timestamps as undefined behavior. I'm not sure one would send multiline strings here and expect some determined results. Perhaps, we could add a validation for a single-line value instead for these types.

jhass: I wouldn't even discuss as much what makes sense or not on the protocol level, it should be a coding style rule to use `\A`/`\z`unless one explicitly wants the line-wise behavior.

cmrd-senya: opened #47

SuperTux88: I agree with @jhass here, it's better codestyle and I also used `\A`/`\z` at other places, but not everywhere (so it's my fault), so we should do it everywhere unless we need the line behavior. :) Thanks for your PR.

PR #47, opened 2017-01-04
Replace ^ and $ with \A \z in patterns

cmrd-senya:
One should use \A and \z unless explicitly wants the line-wise behavior.


SuperTux88:
Thanks 👍

PR #48, opened 2017-01-05
Use enriched_properties instead of properties in Entity#to_h

cmrd-senya:
By the results of the discussion at #42, there are changes to the `to_h` method.

Since `to_h` is intensively used for data generation for entity initializers in the `diaspora_federation` test suite, a move from `properties` to `enriched_properties` requires to uptade entity construction routine to support initialization of properties (boolean, integer, time) with string values.


cmrd-senya:
I merged your changes into my branch and squashed everything into a single commit. Is there anything else I can do here?

SuperTux88:
That's perfect, thanks! 👍

PR #51, opened 2017-01-16
Add a link to the source code

Flaburgan:
Not obvious to know that the code is hosted at https://github.com/diaspora/diaspora_federation from the doc ;)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -17,5 +17,14 @@ This document specifies the future protocol for diaspora. diaspora\* release 0.6
 [entities][entities] with this protocol, but still sends entities with an older protocol. Starting with diaspora\*
 release X this protocol is fully supported.
 
+## Implementations
+
+An implementation of this protocol is available as a Ruby Gem under the AGPL [on Github][github]. This is the library used by the diaspora* project.
+
+The [Friendica][friendica] project also has [its implementation in PHP][phpimplementation].
+
 [communication]: https://wiki.diasporafoundation.org/How_we_communicate
 [entities]: {{ site.baseurl }}/entities/
+[github]: https://github.com/diaspora/diaspora_federation
+[friendica]: https://github.com/friendica/friendica
+[phpimplementation]: https://github.com/friendica/friendica/blob/master/include/diaspora.php


docs/index.md
```diff
@@ -19,3 +19,7 @@ release X this protocol is fully supported.
 
 [communication]: https://wiki.diasporafoundation.org/How_we_communicate
 [entities]: {{ site.baseurl }}/entities/
+
+## Implementation
+
+An implementation of this protocol is available as a Ruby on Rails Gem under the AGPL [on github](https://github.com/diaspora/diaspora_federation)
```
SuperTux88: "Ruby on Rails Gem" -> "a Ruby gem" (explanation: Most of the code is generic ruby in the gem `diaspora_federation` and only a very small part is rails specific in the gem `diaspora_federation-rails`. You can use the code without rails if you write the controllers yourself.)

docs/index.md
```diff
@@ -19,3 +19,7 @@ release X this protocol is fully supported.
 
 [communication]: https://wiki.diasporafoundation.org/How_we_communicate
 [entities]: {{ site.baseurl }}/entities/
```
SuperTux88: Can you add your block above line 20 and add the link to github to this list:
```[github]: https://github.com/diaspora/diaspora_federation```
and use it with `[on Github][github]` in your text.

**Comment by SuperTux88** [comment]
Just an idea: maybe we should also mention other projects using this protocol? I know that friendica is fully compatible with it, I don't know the exact state of redmatrix and hubzilla (if they are compatible with the new version of the protocol).

**Comment by Flaburgan** [comment]
Fixed. Pinging @annando, is there a specific place where we could link to the PHP implementation of the federation?

**Comment by annando** [comment]
@Flaburgan The direct link to the Friendica implementation of the Diaspora protocol is here: https://github.com/friendica/friendica/blob/master/include/diaspora.php

**Comment by Flaburgan** [comment]
@SuperTux88 which link do you prefer?

**Comment by SuperTux88** [comment]
I don't know, maybe we can also link both, and link the "implementation in PHP" to the actual implementation.

**Comment by Flaburgan** [comment]
Done.

**Comment by SuperTux88** [comment]
Thanks ❤️

**Comment by cmrd-senya** [comment]
[There is](https://github.com/jaywink/federation) also a Python implementation from @jaywink. Should we add it here as well?

**Comment by SuperTux88** [comment]
I don't know how complete this implementation is and if it is compatible with the changes.

File: docs/index.md [hunk 3]
```diff
@@ -17,5 +17,14 @@ This document specifies the future protocol for diaspora. diaspora\* release 0.6
...
 release X this protocol is fully supported.
 
+## Implementations
+
+An implementation of this protocol is available as a Ruby Gem under the AGPL [on Github][github]. This is the library used by the diaspora* project.
+
+The [Friendica][friendica] project also has [its implementation in PHP][phpimplementation].
+
 [communication]: https://wiki.diasporafoundation.org/How_we_communicate
 [entities]: {{ site.baseurl }}/entities/
```
cmrd-senya: This document doesn't get rendered well at the github MD viewer:


(Look at the bottom line)

cmrd-senya:
> I don't know how complete this implementation is and if it is compatible with the changes.

[The docs](https://federation.readthedocs.io/en/latest/protocols.html#diaspora) say:

> Note! Diaspora project is currently rewriting parts of the protocol. This library aims to support the new version. When possible, compatibility will be kept with the current and legacy versions but this is not the main objective.

The supported features are listed at the same page. Of course, we can ask @jaywink directly for if we have specific questions.

docs/index.md
cmrd-senya: Ok, I didn't get this from the beginning. Now I see that the main view point for the doc is [this page](https://diaspora.github.io/diaspora_federation/). So it is not expected to work well out from github.

SuperTux88: Yes, the docs are rendered on [github pages](https://pages.github.com/), you can test this locally with `gem install jekyll` and `jekyll serve`.

SuperTux88:
I think it's the best if @jaywink decides that, and if he thinks it should be added here, he can create a PR with a short description and the link to be added (or tell us the details, then we can add it).

Flaburgan:
The best solution would be to allow other implementations to be tested with our test suite. Only those with a full green would be listed.

cmrd-senya:
> The best solution would be to allow other implementations to be tested with our test suite. Only those with a full green would be listed.

Yes, but developing a test suite that supports testing some "random" implementation is not straightforward. It would require some extra efforts. We would need to describe an common interface so that an implementation could communicate with the test suite. At least the test suite should be able to tell the implementation to run a specific action (e.g. send a sharing request).

jaywink:
Thanks @cmrd-senya for the ping :) I'll create a PR 👍 The Python federation library currently has partial support, but I'm not sure that should be a stopping issue in mentioning what libraries implement the protocol. Seems like common sense to me to build a healthy ecosystem for the protocol to aim for adoption. Some libraries might not want to even implement everything, simply because they might see something as unnecessary, for example.

My library is built looking at the new docs, so sometimes after deploying I've had to introduce backwards compatibility for the current stuff as pods don't support the new stuff yet. It's used at least in two projects, [Social-Relay](https://github.com/jaywink/social-relay) and [Socialhome](https://github.com/jaywink/socialhome), both which have running real servers handling federation traffic.

> The best solution would be to allow other implementations to be tested with our test suite. Only those with a full green would be listed.

I'm not really sure what problem this would solve, considering especially the amount of work such a test system would take to set up and keep available. It's very unlikely that the internet will fill itself with libraries adopting the protocol - it would likely be a positive problem ;)

The best thing IMHO would be for the documentation to help developers find resources, be they 100% implementations (at any given time - that degrades anyway), partial implementations or just tools/etc - anything that makes the ecosystem powered by the Diaspora protocol better.

PR #52, opened 2017-01-17
Entity#to_json/.from_json methods

cmrd-senya:
Currently there are only specifications of the `Entity#to_json` and `Entity.from_json` there. Early PR to discuss the details. Implementation of the methods will follow.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,9 @@ group :test do
   # test helpers
   gem "factory_girl_rails",        "~> 4.7"
   gem "fixture_builder",           "0.5.0"
+  gem "json-schema-rspec", "0.0.4"
   gem "rspec-collection_matchers", "~> 1.1.2"
+  gem "rspec-json_expectations",   "~> 2.1"
   gem "webmock",                   "~> 2.0"
 end
 
--- a/lib/diaspora_federation/entities/relayable.rb
+++ b/lib/diaspora_federation/entities/relayable.rb
@@ -14,9 +14,9 @@ module DiasporaFederation
       # @return [Array] order from xml
       attr_reader :xml_order
 
-      # Additional properties from parsed xml
-      # @return [Hash] additional xml elements
-      attr_reader :additional_xml_elements
+      # Additional properties from parsed input object
+      # @return [Hash] additional elements
+      attr_reader :additional_data
 
       # On inclusion of this module the required properties for a relayable are added to the object that includes it.
       #
@@ -62,18 +62,18 @@ module DiasporaFederation
           entity :parent, Entities::RelatedEntity
         end
 
-        klass.extend ParseXML
+        klass.extend Parsing
       end
 
       # Initializes a new relayable Entity with order and additional xml elements
       #
       # @param [Hash] data entity data
       # @param [Array] xml_order order from xml
-      # @param [Hash] additional_xml_elements additional xml elements
+      # @param [Hash] additional_data additional xml elements
       # @see DiasporaFederation::Entity#initialize
-      def initialize(data, xml_order=nil, additional_xml_elements={})
+      def initialize(data, xml_order=nil, additional_data={})
         @xml_order = xml_order.reject {|name| name =~ /signature/ } if xml_order
-        @additional_xml_elements = additional_xml_elements
+        @additional_data = additional_data
 
         super(data)
       end
@@ -94,6 +94,15 @@ module DiasporaFederation
       # @return [String] string representation of this object
       def to_s
         "#{super}#{":#{parent_type}" if respond_to?(:parent_type)}:#{parent_guid}"
+      end
+
+      def to_json
+        super.merge!(property_order: signature_order).tap {|json_hash|
+          missing_properties = json_hash[:property_order] - json_hash[:entity_data].keys
+          missing_properties.each {|property|
+            json_hash[:entity_data][property] = nil
+          }
+        }
       end
 
       private
@@ -151,7 +160,7 @@ module DiasporaFederation
       #
       # @return [Hash] properties with updated signatures
       def enriched_properties
-        super.merge(additional_xml_elements).tap do |hash|
+        super.merge(additional_data).tap do |hash|
           hash[:author_signature] = author_signature || sign_with_author
           hash[:parent_author_signature] = parent_author_signature || sign_with_parent_author_if_available.to_s
         end
@@ -179,40 +188,31 @@ module DiasporaFederation
 
       # @return [String] signature data string
       def signature_data
-        data = normalized_properties.merge(additional_xml_elements)
+        data = normalized_properties.merge(additional_data)
         signature_order.map {|name| data[name] }.join(";")
       end
 
-      # Override class methods from {Entity} to parse the xml
-      module ParseXML
-        private
+      # Override class methods from {Entity} to parse serialized data
+      module Parsing
+        # Does the same job as Entity.from_hash except of the following differences:
+        # 1) unknown properties from the properties_hash are stored to additional_data of the relayable instance
+        # 2) parent entity fetch is attempted
+        # 3) signatures verification is performed; property_order is used as the order in which properties are composed
+        # to compute signatures
+        # 4) unknown properties' keys must be of String type
+        #
+        # @see Entity.from_hash
+        def from_hash(properties_hash, property_order)
+          # Use all known properties to build the Entity (entity_data). All additional elements
+          # are respected and attached to a hash as string (additional_data). This is needed
+          # to support receiving objects from the future versions of diaspora*, where new elements may have been added.
+          additional_data = properties_hash.reject {|key, _| class_props.has_key?(key) }
 
-        # @param [Nokogiri::XML::Element] root_node xml nodes
-        # @return [Entity] instance
-        def populate_entity(root_node)
-          # Use all known properties to build the Entity (entity_data). All additional xml elements
-          # are respected and attached to a hash as string (additional_xml_elements). It also remembers
-          # the order of the xml-nodes (xml_order). This is needed to support receiving objects from
-          # the future versions of diaspora*, where new elements may have been added.
-          entity_data = {}
-          additional_xml_elements = {}
-
-          xml_order = root_node.element_children.map do |child|
-            xml_name = child.name
-            property = find_property_for_xml_name(xml_name)
-
-            if property
-              entity_data[property] = parse_element_from_node(xml_name, class_props[property], root_node)
-              property
-            else
-              additional_xml_elements[xml_name] = child.text
-              xml_name
-            end
-          end
-
-          fetch_parent(entity_data)
-          new(entity_data, xml_order, additional_xml_elements).tap(&:verify_signatures)
+          fetch_parent(properties_hash)
+          new(properties_hash, property_order, additional_data).tap(&:verify_signatures)
         end
+
+        private
 
         def fetch_parent(data)
           type = data.fetch(:parent_type) {
@@ -231,6 +231,14 @@ module DiasporaFederation
           # Fetch and receive parent from remote, if not available locally
           Federation::Fetcher.fetch_public(data[:author], type, guid)
           data[:parent] = DiasporaFederation.callbacks.trigger(:fetch_related_entity, type, guid)
+        end
+
+        def xml_parser_class
+          DiasporaFederation::Parsers::RelayableXmlParser
+        end
+
+        def json_parser_class
+          DiasporaFederation::Parsers::RelayableJsonParser
         end
       end
 
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -109,10 +109,21 @@ module DiasporaFederation
     # @param [Nokogiri::XML::Element] root_node xml nodes
     # @return [Entity] instance
     def self.from_xml(root_node)
-      raise ArgumentError, "only Nokogiri::XML::Element allowed" unless root_node.instance_of?(Nokogiri::XML::Element)
-      raise InvalidRootNode, "'#{root_node.name}' can't be parsed by #{name}" unless root_node.name == entity_name
+      from_hash(*xml_parser_class.new(self).parse(root_node))
+    end
 
-      populate_entity(root_node)
+    private_class_method def self.xml_parser_class
+      DiasporaFederation::Parsers::XmlParser
+    end
+
+    # Creates an instance of self by parsing a hash in the format of JSON serialized object (which usually means
+    # data from a parsed JSON input).
+    def self.from_json(json_hash)
+      from_hash(*json_parser_class.new(self).parse(json_hash))
+    end
+
+    private_class_method def self.json_parser_class
+      DiasporaFederation::Parsers::JsonParser
     end
 
     # Makes an underscored, lowercase form of the class name
@@ -147,6 +158,33 @@ module DiasporaFederation
     # @return [String] string representation of this object
     def to_s
       "#{self.class.name.rpartition('::').last}#{":#{guid}" if respond_to?(:guid)}"
+    end
+
+    # Renders entity to a hash representation of the entity JSON format
+    # @return [Hash] Returns a hash that is equal by structure to the entity in JSON format
+    def to_json
+      {
+        entity_type: self.class.entity_name,
+        entity_data: json_data
+      }
+    end
+
+    # Creates an instance of self, filling it with data from a provided hash of properties.
+    #
+    # The hash format is described as following:<br>
+    # 1) Properties of the hash are representation of the entity's class properties<br>
+    # 2) Keys of the hash must be of Symbol type<br>
+    # 3) Possible values of the hash properties depend on the types of the entity's class properties<br>
+    # 4) Basic properties, such as booleans, strings, integers and timestamps are represented by values of respective
+    # formats<br>
+    # 5) Nested hashes and arrays of hashes are allowed to represent nested entities. Nested hashes follow the same
+    # format as the parent hash.<br>
+    # 6) Besides, the nested entities can be passed in the hash as already instantiated objects of the respective type.
+    #
+    # @param [Hash] properties_hash A hash of the expected format
+    # @return [Entity] an instance
+    def self.from_hash(properties_hash)
+      new(properties_hash)
     end
 
     private
@@ -225,7 +263,7 @@ module DiasporaFederation
 
     def normalize_property(name, value)
       case self.class.class_props[name]
-      when :string, :integer, :boolean
+      when :string
         value.to_s
       when :timestamp
         value.nil? ? "" : value.utc.iso8601
@@ -245,8 +283,8 @@ module DiasporaFederation
     end
 
     def add_property_to_xml(doc, root_element, name, value)
-      if value.is_a? String
-        root_element << simple_node(doc, name, value)
+      if [String, TrueClass, FalseClass, Integer].any? {|c| value.is_a? c }
+        root_element << simple_node(doc, name, value.to_s)
       else
         # call #to_xml for each item and append to root
         [*value].compact.each do |item|
@@ -264,94 +302,27 @@ module DiasporaFederation
       end
     end
 
-    # @param [Nokogiri::XML::Element] root_node xml nodes
-    # @return [Entity] instance
-    private_class_method def self.populate_entity(root_node)
-      new(entity_data(root_node))
-    end
+    # Generates a hash with entity properties which is put to the "entity_data"
+    # field of a JSON serialized object.
+    # @return [Hash] object properties in JSON format
+    def json_data
+      enriched_properties.map {|key, value|
+        type = self.class.class_props[key]
 
-    # @param [Nokogiri::XML::Element] root_node xml nodes
-    # @return [Hash] entity data
-    private_class_method def self.entity_data(root_node)
-      class_props.map {|name, type|
-        value = parse_element_from_node(name, type, root_node)
-        [name, value] unless value.nil?
-      }.compact.to_h
-    end
-
-    # @param [String] name property name to parse
-    # @param [Class, Symbol] type target type to parse
-    # @param [Nokogiri::XML::Element] root_node XML node to parse
-    # @return [Object] parsed data
-    private_class_method def self.parse_element_from_node(name, type, root_node)
-      if type.instance_of?(Symbol)
-        parse_string_from_node(name, type, root_node)
-      elsif type.instance_of?(Array)
-        parse_array_from_node(type.first, root_node)
-      elsif type.ancestors.include?(Entity)
-        parse_entity_from_node(type, root_node)
-      end
-    end
-
-    # Create simple entry in data hash
-    #
-    # @param [String] name xml tag to parse
-    # @param [Class, Symbol] type target type to parse
-    # @param [Nokogiri::XML::Element] root_node XML root_node to parse
-    # @return [String] data
-    private_class_method def self.parse_string_from_node(name, type, root_node)
-      node = root_node.xpath(name.to_s)
-      node = root_node.xpath(xml_names[name].to_s) if node.empty?
-      parse_string(type, node.first.text) if node.any?
-    end
-
-    # @param [Symbol] type target type to parse
-    # @param [String] text data as string
-    # @return [String, Boolean, Integer, Time] data
-    private_class_method def self.parse_string(type, text)
-      case type
-      when :timestamp
-        begin
-          Time.parse(text).utc
-        rescue
-          nil
+        if !value.nil? && type.instance_of?(Class) && value.respond_to?(:to_json)
+          entity_data = value.to_json
+          [key, entity_data] unless entity_data.nil?
+        elsif type.instance_of?(Array)
+          entity_data = value.nil? ? nil : value.map(&:to_json)
+          [key, entity_data] unless entity_data.nil?
+        else
+          [key, value]
         end
-      when :integer
-        text.to_i if text =~ /\A\d+\z/
-      when :boolean
-        return true if text =~ /\A(true|t|yes|y|1)\z/i
-        false if text =~ /\A(false|f|no|n|0)\z/i
-      else
-        text
-      end
-    end
-
-    # Create an entry in the data hash for the nested entity
-    #
-    # @param [Class] type target type to parse
-    # @param [Nokogiri::XML::Element] root_node XML node to parse
-    # @return [Entity] parsed child entity
-    private_class_method def self.parse_entity_from_node(type, root_node)
-      node = root_node.xpath(type.entity_name)
-      type.from_xml(node.first) if node.any? && node.first.children.any?
-    end
-
-    # Collect all nested children of that type and create an array in the data hash
-    #
-    # @param [Class] type target type to parse
-    # @param [Nokogiri::XML::Element] root_node XML node to parse
-    # @return [Array<Entity>] array with parsed child entities
-    private_class_method def self.parse_array_from_node(type, root_node)
-      node = root_node.xpath(type.entity_name)
-      node.select {|child| child.children.any? }.map {|child| type.from_xml(child) } unless node.empty?
+      }.compact.to_h
     end
 
     # Raised, if entity is not valid
     class ValidationError < RuntimeError
-    end
-
-    # Raised, if the root node doesn't match the class name
-    class InvalidRootNode < RuntimeError
     end
 
     # Raised, if the entity name in the XML is invalid
--- a/lib/diaspora_federation/parsers/json_parser.rb
+++ b/lib/diaspora_federation/parsers/json_parser.rb
@@ -1,0 +1,60 @@
+module DiasporaFederation
+  module Parsers
+    # This is a parser of JSON serialized object. JSON object format is defined by
+    # JSON schema which is available at https://diaspora.github.io/diaspora_federation/schemas/federation_entities.json.
+    # TODO: We must publish the schema at a real URL
+    class JsonParser < BaseParser
+      # @see BaseParser#parse
+      # @param [Hash] json_hash A hash acquired by running JSON.parse with JSON serialized entity
+      # @return [Array[1]] comprehensive data for an entity instantiation
+      def parse(json_hash)
+        from_json_sanity_validation(json_hash)
+        parse_entity_data(json_hash["entity_data"])
+      end
+
+      private
+
+      def parse_entity_data(entity_data)
+        hash = entity_data.map {|key, value|
+          property = entity_type.find_property_for_xml_name(key)
+          if property
+            type = entity_type.class_props[property]
+            [property, parse_element_from_value(type, entity_data[key])]
+          else
+            [key, value]
+          end
+        }.to_h
+
+        [hash]
+      end
+
+      def parse_element_from_value(type, value)
+        return if value.nil?
+        if %i(integer boolean timestamp).include?(type) && !value.is_a?(String)
+          value
+        elsif type.instance_of?(Symbol)
+          parse_string(type, value)
+        elsif type.instance_of?(Array)
+          raise DeserializationError, "Expected array for #{type}" unless value.respond_to?(:map)
+          value.map {|element|
+            type.first.from_json(element)
+          }
+        elsif type.ancestors.include?(Entity)
+          type.from_json(value)
+        end
+      end
+
+      def from_json_sanity_validation(json_hash)
+        missing = %w(entity_type entity_data).map {|prop|
+          prop if json_hash[prop].nil?
+        }.compact.join(", ")
+        raise DeserializationError, "Required properties are missing in JSON object: #{missing}" unless missing.empty?
+        assert_parsability_of(json_hash["entity_type"])
+      end
+
+      # Raised when the format of the input JSON data doesn't match the parser's expectations
+      class DeserializationError < RuntimeError
+      end
+    end
+  end
+end
--- a/lib/diaspora_federation/parsers/relayable_json_parser.rb
+++ b/lib/diaspora_federation/parsers/relayable_json_parser.rb
@@ -1,0 +1,25 @@
+module DiasporaFederation
+  module Parsers
+    # This is a parser of JSON serialized object, that is normally used for parsing data of relayables.
+    # Assumed format differs from the usual entity by additional "property_order" property which is used to
+    # compute signatures deterministically.
+    # Input JSON for this parser is expected to match "/definitions/relayable" subschema of the JSON schema at
+    # https://diaspora.github.io/diaspora_federation/schemas/federation_entities.json.
+    class RelayableJsonParser < JsonParser
+      # @see JsonParser#parse
+      # @see BaseParser#parse
+      # @return [Array[2]] comprehensive data for an entity instantiation
+      def parse(json_hash)
+        super.push(json_hash["property_order"])
+      end
+
+      private
+
+      def from_json_sanity_validation(json_hash)
+        super
+        return unless json_hash["property_order"].nil?
+        raise DeserializationError, "Required property is missing in JSON object: property_order"
+      end
+    end
+  end
+end
--- a/lib/diaspora_federation/schemas/federation_entities.json
+++ b/lib/diaspora_federation/schemas/federation_entities.json
@@ -1,0 +1,373 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://diaspora.github.io/diaspora_federation/schemas/federation_entities-0.2.0-dev.json",
+  "oneOf": [
+    {"$ref": "#/definitions/comment"},
+    {"$ref": "#/definitions/like"},
+    {"$ref": "#/definitions/participation"},
+    {"$ref": "#/definitions/poll_participation"},
+    {"$ref": "#/definitions/status_message"},
+    {"$ref": "#/definitions/reshare"},
+    {"$ref": "#/definitions/profile"},
+    {"$ref": "#/definitions/location"},
+    {"$ref": "#/definitions/photo"},
+    {"$ref": "#/definitions/poll"},
+    {"$ref": "#/definitions/poll_answer"}
+  ],
+
+  "definitions": {
+    "signature": {
+      "type": "string",
+      "minLength": 30
+    },
+
+    "guid": {
+      "type": "string",
+      "minLength": 16,
+      "maxLength": 255
+    },
+
+    "relayable": {
+      "type": "object",
+      "description": "please don't use this object unless you're defining a new child relayable schema",
+      "properties": {
+        "entity_data": {
+          "type": "object",
+          "properties": {
+            "author": { "type": "string" },
+            "guid": { "$ref": "#/definitions/guid" },
+            "parent_guid": { "$ref": "#/definitions/guid" },
+            "author_signature": { "$ref": "#/definitions/signature" },
+            "parent_author_signature": { "$ref": "#/definitions/signature" }
+          },
+          "required": [
+            "author", "guid", "parent_guid"
+          ]
+        },
+        "property_order": {
+          "type": "array",
+          "items": { "type": "string" }
+        }
+      },
+      "required": [
+        "entity_data", "entity_type", "property_order"
+      ]
+    },
+
+    "comment": {
+      "allOf": [
+        {"$ref": "#/definitions/relayable"},
+        {
+          "type": "object",
+          "properties": {
+            "entity_data": {
+              "type": "object",
+              "properties": {
+                "text": { "type": "string" },
+                "created_at": { "type": "string" }
+              },
+              "required": ["text"]
+            },
+            "entity_type": {
+              "type": "string",
+              "pattern": "^comment$"
+            }
+          }
+        }
+      ]
+    },
+
+    "like": {
+      "allOf": [
+        {"$ref": "#/definitions/relayable"},
+        {
+          "type": "object",
+          "properties": {
+            "entity_type": {
+              "type": "string",
+              "pattern": "^like$"
+            },
+            "entity_data": {
+              "type": "object",
+              "properties": {
+                "positive": { "type": "boolean" },
+                "parent_type": { "enum": ["Post", "Comment"] }
+              },
+              "required": ["positive"]
+            }
+          }
+        }
+      ]
+    },
+
+    "participation": {
+      "allOf": [
+        {"$ref": "#/definitions/relayable"},
+        {
+          "type": "object",
+          "properties": {
+            "entity_type": {
+              "type": "string",
+              "pattern": "^participation$"
+            },
+            "entity_data": {
+              "type": "object",
+              "properties": {
+                "parent_type": {"enum": ["Post"]}
+              }
+            }
+          }
+        }
+      ]
+    },
+
+    "poll_participation": {
+      "allOf": [
+        {"$ref": "#/definitions/relayable"},
+        {
+          "type": "object",
+          "properties": {
+            "entity_type": {
+              "type": "string",
+              "pattern": "^poll_participation$"
+            },
+            "entity_data": {
+              "type": "object",
+              "properties": {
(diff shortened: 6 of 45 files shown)


lib/diaspora_federation/entity.rb
```diff
@@ -149,6 +149,19 @@ def to_s
...
     end
 
+    # Might be used to modify entity JSON object just before serialization
+    # @return [Hash] Returns a hash that is equal by structure to the entity in JSON format
+    def to_json_hash
+      {}
+    end
+
+    # @return [String] Renders the entity to the JSON representation
+    delegate :to_json, to: :to_json_hash
```
cmrd-senya: I added this because of Rubocop ranting. If thi is not desirable, perhaps, we should change it's settings.

SuperTux88: I just disabled the cop for lib/* (and updated rubocop) in the develop branch. So it shouldn't rant after your next rebase :)

cmrd-senya:
> Just some quick thoughts: I think that the structure of relayables and and other entities should be more similar. So I would make entity_data like signable_data so you can name both the same and parse the same way. An alternative would be to make relayables also like entity_data and add an entity_order field. What do you think?

We must preserve the order of the fields somehow, that's why I used "array of hashes" approach. Alternatively, we can make a similar `entity_data` hash for relayables. In that case we should have the order stored separately. Array of strings with the property names is one of the optimal solutions. I can respecify it this way, if you see it apropriate. Also, I think `property_order` is a bit more meaningful name for that than `entity_order`.

Are there any more changes we want for the design of the JSON de/serialization feature?

lib/diaspora_federation/entity.rb
cmrd-senya: I don't see any new commits after #53 merge in https://github.com/diaspora/diaspora_federation/commits/develop.

SuperTux88:
Yes, I just think we should make it somewhat consistent over all entities, and since "array of hashes" feels weird with normal entities I think a seperate order field is OK (we also have a seperate `entity_class` field compared to xml). I think `property_order` is OK.

And I would move `author_signature` and `parent_author_signature` also inside `entity_data`, so that everything that is defined as property in the Entity is included there (like in xml). So you have `entity_class`, `entity_data` (with all the properties) and for relayables `property_order` for the order. An alternative would be to create `signature_data` in addition to `entity_data` and put `property_order`, `author_signature` and `parent_author_signature` there, but I think that would need more special handling in parse and generate. What do you think?

I don't know much about json schemas (probably need to learn something here ;) ), so I don't know if you got this right, but keep in mind that every entity can have additional unknown properties (which are ignored when parsing, but don't make the json invalid). Also some fields are optional.

lib/diaspora_federation/entity.rb
SuperTux88: Oh, I didn't see push failed before, sorry ... now it's pushed :)

cmrd-senya:
> And I would move author_signature and parent_author_signature also inside entity_data, so that everything that is defined as property in the Entity is included there (like in xml). So you have entity_class, entity_data (with all the properties) and for relayables property_order for the order. An alternative would be to create signature_data in addition to entity_data and put property_order, author_signature and parent_author_signature there, but I think that would need more special handling in parse and generate. What do you think?

Well, yeah, if we put signatures along with other properties to the `entity_data`, then the implementation will be at most close to the XML implementation, which likely makes it simpler. I'll give it another look, to check if there are any strong reasons to implement it differently.

> I don't know if you got this right, but keep in mind that every entity can have additional unknown properties (which are ignored when parsing, but don't make the json invalid). Also some fields are optional.

By default JSON schema for objects allows any unknown properties in these objects. It is possible to disable properties if they aren't presented in the object description, but that should be stated explicitly using [additionalProperties](http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.18) keyword.

The same is true for fields optionality. All fields are optional by default, unless you specify them in the `"required"` section, like [here](https://github.com/diaspora/diaspora_federation/pull/52/files#diff-04c8b9cac73b7eed0f9ecd351fde9f91R117).

SuperTux88:
> I'll give it another look, to check if there are any strong reasons to implement it differently.

Yes, I think it should be as close to xml as possible (unless there are good reasons). 👍 

And thanks for the explanation about JSON schema ❤️

cmrd-senya:
> And thanks for the explanation about JSON schema :heart:

Feel free to ask anything else on JSON schemas. It's not 100% that I'll know the answer, but I've learnt some concepts already :)

cmrd-senya:
Can I ask your for another review round please?

Here I have an implementation of to/from_json methods, but it is still draft and I'm going to polish and clean it some more. However the API methods I introduce and the specifications are more or less what I expect them to be. So I think I need your opinion before I put some more efforts.

cmrd-senya:
Please tell me if you didn't miss my review request.

spec/lib/diaspora_federation/entity_spec.rb
```diff
@@ -268,6 +272,125 @@ module DiasporaFederation
...
+
+      include_examples "it raises error when the entity class doesn't match the entity_class property", <<-JSON
+{
+  "entity_class": "unknown_entity",
+  "entity_data": {}
+}
+JSON
+
+      include_examples ".from_json parse error", "entity_class is missing", '{"entity_data": {}}'
+      include_examples ".from_json parse error", "entity_data is missing", '{"entity_class": "test_complex_entity"}'
```
SuperTux88: Same as above for "to json" here, mixing of testing `.from_json` and `.from_json_hash`.

lib/diaspora_federation/entity.rb
```diff
@@ -149,6 +150,56 @@ def to_s
...
+      return if properties_hash.nil?
+
+      populate_entity {|name, type|
+        parse_element_from_value(type, properties_hash[name.to_s])
+      }
+    end
+
+    # Creates an instance of self by parsing a hash in the format of JSON serialized object (which usually means
+    # data from a parsed JSON input).
+    def self.from_json_hash(json_hash)
```
SuperTux88: I think this can be made a `private_class_method`.

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -185,6 +195,40 @@ def signature_data
...
+        end
+
+        # Does the same job as Entity.from_hash except of the following differences:
+        # 1) unknown properties from the properties_hash are stored to additional_xml_elements of the relayable instance
+        # 2) property_order is used as the order in which properties are composed to compute signatures
+        # 3) parent entity fetch is attempted
+        # 4) signatures verification is performed
+        #
+        # @see Entity.from_hash
+        def from_hash(properties_hash, property_order)
```
SuperTux88: This does basically the same as `populate_entity` below.

lib/diaspora_federation/entity.rb
```diff
@@ -264,21 +315,52 @@ def simple_node(doc, name, value)
...
+    def json_data
+      enriched_properties.map {|key, value|
+        type = self.class.class_props[key]
+
+        if type.instance_of?(Class) && value.respond_to?(:to_json_hash)
+          entity_data = value.to_json_hash[:entity_data]
+          [key, entity_data] unless entity_data.nil?
+        else
+          [key, value]
+        end
```
cmrd-senya: Arrays are properly handled there as a special case of `else` block. Since `json_data` is called in `to_json` which returns a hash, which is later being converted to JSON via `ActiveSupport::ToJsonWithActiveSupportEncoder#to_json` method, array conversion just works with no special handling.

```ruby
[3] pry(#<RSpec::ExampleGroups::DiasporaFederationEntity::ToJson>)> entity_class.new(hash).to_json
=> {:entity_class=>"test_complex_entity",
 :entity_data=>
  {:test1=>"123",
   :test2=>false,
   :test3=>"456",
   :test4=>789,
   :test5=>"2017-02-11T23:52:09Z",
   :test6=>{:test=>"000"},
   :multi=>
    [#<DiasporaFederation::Entities::OtherEntity:0x00000002c4e030 @asdf="01">,
     #<DiasporaFederation::Entities::OtherEntity:0x00000002c4ce10 @asdf="02">]}}
[4] pry(#<RSpec::ExampleGroups::DiasporaFederationEntity::ToJson>)> entity_class.new(hash).to_json.to_json
=> "{\"entity_class\":\"test_complex_entity\",\"entity_data\":{\"test1\":\"123\",\"test2\":false,\"test3\":\"456\",\"test4\":789,\"test5\":\"2017-02-11T23:52:09Z\",\"test6\":{\"test\":\"000\"},\"multi\":[{\"asdf\":\"01\"},{\"asdf\":\"02\"}]}}"
```

spec/lib/diaspora_federation/entities/comment_spec.rb
```diff
@@ -32,6 +32,8 @@ module DiasporaFederation
 
     it_behaves_like "a relayable Entity"
 
+    include_examples "#to_json output matches JSON schema"
```
cmrd-senya: With `include_examples` we have a better full example name:
```
DiasporaFederation::Entities::Comment #to_json output matches JSON schema
```
while with `it_behaves_like` it gives:
```
DiasporaFederation::Entities::Comment behaves like #to_json output matches JSON schema #to_json output matches JSON schema
```
which can be changed to
```
DiasporaFederation::Entities::Comment behaves like #to_json output matches JSON schema
```
if we make `it` argument to be `""` at the shared_example definition, but we still have the "behaves like" words here that doesn't make any sense.

Since we have no technical difference in this case (the context of shared examples doesn't introduce anything, that could affect the context where we do include), I think we should pick the one which makes better name. However, if you insist, I can change it.

lib/diaspora_federation/entity.rb
SuperTux88: Does this mean we need `ActiveSupport` as a dependency now?

spec/lib/diaspora_federation/entities/comment_spec.rb
SuperTux88: Ok, makes sense.

Or we can make it like `it_behaves_like "an XML Entity"`: `it_behaves_like "an JSON Entity"` ?

lib/diaspora_federation/entity.rb
cmrd-senya: In the version that I'm preparing now I followed the idea that `to_json` returns a hash, which should be transformed to JSON by user (like it works for `to_xml` where we get `Nokogiri::XML::Element` and not a String with XML). 

So in this very case we don't need ActiveSupport, unless we wan't to transform actual `to_json` output (a hash) to a string with JSON representation.

spec/lib/diaspora_federation/entities/comment_spec.rb
cmrd-senya: > Or we can make it like it_behaves_like "an XML Entity": it_behaves_like "an JSON Entity" ?

No objections.

cmrd-senya:
Here is another update to the PR.

Here I extraced the parsing code to `Parsers`, who do parsing of XML/JSON and produce hashes. The hashes then go to the `.from_hash` method which instantiates entity. Hash format is described in a comment to the `.from_hash`, but basically the hash format is what entity consturctor has taken as an argument before (so no new hash format invented here).

Some docs and specs are incomplete, but I tried to mark them with a TODO (and a few other questionable places).

I can haz your look on the PR please?

lib/diaspora_federation/entity.rb
```diff
@@ -109,10 +109,29 @@ def to_xml
     # @param [Nokogiri::XML::Element] root_node xml nodes
     # @return [Entity] instance
     def self.from_xml(root_node)
-      raise ArgumentError, "only Nokogiri::XML::Element allowed" unless root_node.instance_of?(Nokogiri::XML::Element)
-      raise InvalidRootNode, "'#{root_node.name}' can't be parsed by #{name}" unless root_node.name == entity_name
+      from_hash(*xml_parser.parse_xml(root_node))
```
SuperTux88: I would just use `xml_parser_class.new(self)` here directly.

cmrd-senya:
>  I'd rather keep .from_hash private (and maybe even keep the name .populate_entity?), because it shouldn't be part of the public API (users of the library should use .new instead).

We have `to_h` public method so having `from_hash` private would be somewhat asymmetric. What are the reasons to keep public `to_h` method? What are possible use cases for it? What justifies the fact that we allow a user requesting a hash for an entity but we disallow user to build an entity from a hash?

If I'm not mistaken (I feel lazy for verifying it right now), the hashes formats are actually compatible with each other so one can do `Entity.from_hash(entity.to_h)`, so it feels like parts of the same interface.

cmrd-senya:
> Technically we don't need the entity_class here, because we can get it from class_props, so it's kind of redundant, but on the other hand we could also validate it.

What validations are you talking about exactly? If you're talking about parser validations ([like this one](https://github.com/diaspora/diaspora_federation/pull/52/files#diff-0e0b79d5b15ea2212f10b1a42c0eff1cR54)) then it does only the verification of expected JSON/XML structure and class_props type match with the class name from the document. Which is not a self-sufficient reason to introduce the `entity_class` property, because it is actually the presence of the `entity_class` is the reason why we need these validations. If I'm not mistaken all other validations happen here no matter if the entity is nested or not.

> In the JSON schema we can't just reference to it, when something exists as it's own entity and nested in another entity (photo for example).

We can define a schema for the `entity_data` object separately and then reference it from both container classes (`StatusMessage` for `Photo`) and full entity classes. Tecnhically it's not a problem. It makes the JSON schema less pretty, but the JSON itself would be shorter and neater. But if we decide not to include `entity_class` for nested objects and then we need to reference this schema from it's own entity, it will be possible.

> We can't nest relayables (Currently not used. But maybe we want to nest comments/likes in the post for fetching or the export? That needs some more work because the current workflow needs the parent to be already persisted, so it is out of scope here, this is already big enough. But who knows, maybe something to think about in the future?)

Again, it is solvable, we could define basing on `class_props` that when it is a relayable we should use the full entity format with `entity_class`, `entity_data` and `property_order`.

So I don't feel that the problems you listed here can't be solved with my present approach, so the present implementation is not a blocker and could be left as is. However doing as you propose will make both our parsing implementation and JSON schema simpler and more straightforward. The only advantage of the approach I've implemented so far (without `entity_class`) is that it has less redundancy in the output document structure. But I can't really think of any strong advantage of having lower redundancy here. What do you think?

lib/diaspora_federation/discovery/web_finger.rb
```diff
@@ -148,7 +148,7 @@ def to_xml
       # @return [WebFinger] WebFinger instance
       # @raise [InvalidData] if the given XML string is invalid or incomplete
       def self.from_xml(webfinger_xml)
-        data = parse_xml_and_validate(webfinger_xml)
+        data = parse_and_validate(webfinger_xml)
```
SuperTux88: I think we should keep that (because it is really "parse xml" here and it is a special case for webfinger), but it's probably unintentionally because of search/replace?

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -179,41 +188,36 @@ def signature_order
...
+        # to compute signatures
+        # 4) unknown properties' keys must be of String type
+        #
+        # @see Entity.from_hash
+        def from_hash(properties_hash, property_order)
+          return if properties_hash.nil?
+          # Use all known properties to build the Entity (entity_data). All additional elements
+          # are respected and attached to a hash as string (additional_data). This is needed
+          # to support receiving objects from the future versions of diaspora*, where new elements may have been added.
+          entity_data = properties_hash
```
SuperTux88: I didn't notice this last time, but I think we don't need this line, when we don't change `properties_hash`. (See comment below)

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -179,41 +188,36 @@ def signature_order
...
+          additional_data = entity_data.keys.each_with_object({}) do |key, hash|
+            hash[key] = entity_data.delete(key) unless class_props.has_key?(key)
```
SuperTux88: Do we need to `delete` it from the `entity_data`? I think the entity-constructor ignores unknown `class_props` anyway.

I think a simple `additional_data = properties_hash.reject {|k,v| class_props.has_key?(k) }` should do it.

spec/support/shared_entity_specs.rb
```diff
@@ -133,3 +133,55 @@ def verify_signature(pubkey, signature, signed_string)
...
   end
 end
+
+shared_examples "a JSON Entity" do
+  it "#to_json output matches JSON schema" do
+    json = described_class.new(data).to_json
+    expect(json.to_json).to match_json_schema(:entity_schema)
+  end
+
+  include_examples "common Entity JSON expectations" do
```
SuperTux88: > I would add that to the `shared_examples "a JSON Entity"`

I thought about just merging the `shared_examples` instead of including it . If we test this in `"a JSON Entity"` for every entity, we don't need to test `"common Entity JSON expectations"` for testing for only `Entity` and `Relayable`.

SuperTux88:
> If I'm not mistaken (I feel lazy for verifying it right now), the hashes formats are actually compatible with each other so one can do `Entity.from_hash(entity.to_h)`, so it feels like parts of the same interface.

You're right, if that works, we can probably keep it. I was just worried that people can try to use it instead of `.new` to create the entities (which wouldn't work for relayables if the hash doesn't contain a signature). And `#to_h` is there to have a simple method to get all data out of an entity, which can be useful anyway, I never thought about to parse it back with `.from_hash`. I still don't know if we need it, but ... why not :)

> We can define a schema for the `entity_data` object separately ...

Yes, we can, but as you mentioned, it isn't as pretty as when referencing the full entity.

> Again, it is solvable, we could define basing on `class_props` that when it is a relayable we should use the full entity format with `entity_class`, `entity_data` and `property_order`.

Everything is "solvable", but please no special rules for relayables where it isn't really needed.

> ... What do you think?

I never said that it is "a blocker", but as you said, I think it makes everything "simpler and more straightforward", with less special cases. The only "disadvantage" is that we have a small redundancy in the output JSON, but I think that we should do the "simpler and more straightforward"-solution, and I think the output JSON format is also more consistent :)

Gemfile
```diff
@@ -51,7 +51,9 @@ group :test do
   # test helpers
   gem "factory_girl_rails",        "~> 4.7"
   gem "fixture_builder",           "0.5.0"
+  gem "json-schema-rspec", "0.0.4"
   gem "rspec-collection_matchers", "~> 1.1.2"
+  gem "rspec-json_expectations", "~> 2.1"
```
SuperTux88: Can you fix the indention here (pronto doesn't see it, because rubocop complains about the line below).

spec/spec_helper.rb
```diff
@@ -17,6 +17,8 @@
 
 require "rspec/rails"
 require "webmock/rspec"
+require "rspec/json_expectations"
+
```
SuperTux88: Please remove this newline (rubocop thinks the newline below is the "extra blank line" ... so pronto doesn't care ;) I don't care if you delete line 21 or line 22 ;) Just make rubocop happy :) ).

lib/diaspora_federation/discovery/web_finger.rb
cmrd-senya: > but it's probably unintentionally because of search/replace?

Yes, that's what happened here.

spec/support/shared_entity_specs.rb
cmrd-senya: Okay, and what do you think of `JSON is parsable with #from_json` below? Should I merge it with `a JSON Entity` as well?

spec/support/shared_entity_specs.rb
```diff
@@ -133,3 +133,74 @@ def verify_signature(pubkey, signature, signed_string)
...
+end
+
+shared_examples ".from_json return object verification" do
+  it "from_json(entity_json).to_json should match entity.to_json" do
+    entity_json = entity.to_json.to_json
+    expect(entity_class.from_json(JSON.parse(entity_json)).to_json.to_json).to eq(entity_json)
+  end
+end
+
+# TODO: move to a shared_parser_specs.rb?
```
SuperTux88: I thought about this, and I think it is cleaner to move it to it's own file. And an extra file for that doesn't hurt us :)

lib/diaspora_federation/entity.rb
SuperTux88: Hmm, I also thought a bit more about this. So we are still relaying on that somebody uses `ActiveSupport` with the resulting hash?

So we handle single nested entities, because we have the `nil?` check anyway. And I think it doesn't hurt when we also call `.to_json` explicitly for nested arrays here, because then we have a clean hash returned here.

cmrd-senya:
@SuperTux88, I received a notification email from this topic with the following text:

> Hmm, I also thought a bit more about this. So we are still relaying on that somebody uses ActiveSupport with the resulting hash?
>
> So we handle single nested entities, because we have the nil? check anyway. And I think it doesn't hurt when we also call .to_json explicitly for nested arrays here, because then we have a clean hash returned here.

But I can't see it anywhere here. Have you deleted it or is it a github bug?

SuperTux88:
> But I can't see it anywhere here. Have you deleted it or is it a github bug?

It's "hidden" because it's "outdated", but this link should bring you there: https://github.com/diaspora/diaspora_federation/pull/52#discussion_r103339640

lib/diaspora_federation/entity.rb
cmrd-senya: Well, there is also a `to_json` method from `JSON::Ext::Generator::GeneratorMethods::Hash/Array` modules, which are independent of Rails. If one doesn't use ActiveSupport here, he can fallback to these methods. However I have no idea of how to test it with our test suite. If we remove ActiveSupport, we can use these replacements as well, so having this call is probably doesn't make us dependent on ActiveSupport.

SuperTux88: We don't "depend" on ActiveSupport, but we depend on something calling `to_json` again on elements in an array (while having hashes for single nested entities). So I don't like the fact, that we still have "raw" entities in the result of `to_json` and not a clean hash.

spec/support/shared_entity_specs.rb
```diff
@@ -133,3 +133,74 @@ def verify_signature(pubkey, signature, signed_string)
...
+      end
+    }
+  end
+
+  include_examples "JSON is parsable with #from_json" do
+    let(:entity_class) { described_class }
+  end
+end
+
+shared_examples "JSON is parsable with #from_json" do
```
cmrd-senya: So you think it's okay that I remove use of this one and the one below from `entity_spec.rb` and `relayable_entity_spec.rb`?

SuperTux88: Yes, because we test this then for each of the entities and relayables. But since in the specs of relayables we already add signatures in the `data`, maybe add some specs to `relayable_entity_spec.rb` that explicitly tests if signatures/order/additional_data are handled correctly, like we also do for `to/from_xml` (some are already there). But this generic "is parsable" should be OK when it is only used in the `a JSON Entity`.

cmrd-senya: I'm not sure we need to test signatures handling with `to/from_json` since it's responsibility of `from_hash` now and we have tests for that. So even `to/from_xml` tests are mostly redundant here. Maybe I miss some tests for `from_hash` though, I should review it before I can say that definetely.

app/schemas/federation_entities.json
```diff
@@ -0,0 +1,366 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://diasporafoundation.org/schemas/federation_entities-1.0.0.json",
```
cmrd-senya: As our schema seem to get stabilized (at least for this PR) we should decide upon publishing it at some real place. Also we should define versioning schema. I think using semver is okay for it. Maybe we should also make `-latest` symlink?

cmrd-senya:
I think most of the specs for `.from_xml` from `entity_spec.rb` can now be moved to the `xml_parser_spec.rb`. @SuperTux88, do you have any objections?

SuperTux88:
Sounds good. 👍

app/schemas/federation_entities.json
SuperTux88: I don't know about that. Your current `id` points to diasporafoundation.org, pinging @denschub about that. An alternative would be to use https://diaspora.github.io/diaspora_federation/schemas.

Also I'm not sure if we want to use a separate versioning for that, or just re-use the same version as the gem?

lib/diaspora_federation/entity.rb
```diff
@@ -109,10 +109,25 @@ def to_xml
...
+    def self.from_json(json_hash)
+      from_hash(*json_parser.parse(json_hash))
+    end
 
-      populate_entity(root_node)
+    private_class_method def self.json_parser_class
+      DiasporaFederation::Parsers::JsonParser
+    end
+
+    def self.json_parser
```
SuperTux88: This method can now be inlined too, like in `from_xml`.

app/schemas/federation_entities.json
cmrd-senya: > An alternative would be to use https://diaspora.github.io/diaspora_federation/schemas.

The bad side of using `github.io` is that we depend on GitHub. It is recommended to place schemas within the domain of control. I know that chances are low that GitHub will fake our schemas or we'll experience problems because of some kind of outage. However GitHub experienced [some kinds of attacks in the past](https://en.wikipedia.org/wiki/Censorship_of_GitHub). I don't know if `github.io` suffered of any of this, but that in theory is something that can affect diaspora as a user. It doesn't mean that we're completely safe when using our own domain, but it's better to not depend on a different organization if we have a possibility.

> Also I'm not sure if we want to use a separate versioning for that, or just re-use the same version as the gem?

What do we do if the gem is updated but schema isn't? If yes then we'll have a few version numbers matching the same schema. If not, then what do we do when we need to make a patch modification to schema, while gem is changing minor or major version (according to semver)? I'm not sure we need to couple the version indexes here, because schemas are not that strictly tied to the gem's code. There are even such situations possible, when we change the schema, but don't touch gem implementation at all (e.g. adding another constraint to schema which already presents in implementation). Maybe we should even place the schemas under a different repository and even them in a separate gem?

cmrd-senya:
Okay, I looked through the code once again. I think that besides the question of schemas placement everything looks good to me.

app/schemas/federation_entities.json
SuperTux88: > The bad side of using github.io is that we depend on GitHub.

I don't think that that is a problem, we depend on GitHub anyway, and I think when something happens, then the schemes is our last problem. And a user of diaspora shouldn't even notice it, if there is a problem with the schemes on GitHub.

> What do we do if the gem is updated but schema isn't?

We can just change the schema version when it changes (to the gem version with which the new schema is released), and if the schema doesn't change, keep it on the last version where it changed. That would result in versions with gaps, but I don't think that is a problem.

If you really think it is worth to track it's own version for the schemes then I think it's the best to extract it to it's own project and add it as dependency.

But anyway: I'm interested in @denschub opinion on this topic.

spec/lib/diaspora_federation/entities/relayable_spec.rb
```diff
@@ -271,109 +252,283 @@ def verify_signature(pubkey, signature, signed_string)
...
-        expect_callback(:fetch_related_entity, "Parent", parent_guid).and_return(remote_parent)
-        expect(Federation::Fetcher).not_to receive(:fetch_public)
+      context "parsing" do
+        before do
+          expect_callback(:fetch_related_entity, "Parent", parent_guid).and_return(remote_parent)
+          expect_callback(:fetch_public_key, author).and_return(author_pkey.public_key)
+          expect_callback(:fetch_public_key, remote_parent.author).and_return(parent_pkey.public_key)
+        end
+
+        context "when JSON properties are sorted and there is an unknown property" do
```
SuperTux88: Nit: Since this is `.from_hash` (and not JSON), maybe simply write `"when properties ..."`?

spec/lib/diaspora_federation/entities/relayable_spec.rb
```diff
@@ -271,109 +252,283 @@ def verify_signature(pubkey, signature, signed_string)
...
+                parent:                  remote_parent
+              }.merge("new_property" => new_property),
+              %w(author guid parent_guid new_property property),
+              "new_property" => new_property
+            ).and_call_original
+            Entities::SomeRelayable.from_hash(entity_data, property_order)
+          end
+        end
+
+        it "creates Entity with empty 'additional_data' if the xml has only known properties" do
```
SuperTux88: Nit: `... if it has ...` (no "xml")

spec/support/shared_entity_specs.rb
```diff
@@ -133,3 +133,58 @@ def verify_signature(pubkey, signature, signed_string)
...
+    expect(json.to_json).to match_json_schema(:entity_schema)
+  end
+
+  let(:to_json_output) { described_class.new(data).to_json.to_json }
+
+  it "contains described_class property matching the entity class (underscored)" do
+    expect(to_json_output).to include_json(entity_type: described_class.entity_name)
+  end
+
+  it "contains JSON properties for each of the entity properties with the entity_data property" do
```
cmrd-senya: Maybe we should add a test for `from_json` that tests that the input from `let(:json)` is parsed with no exceptions raised instead of just dropping `let(:json)`?

Also I think `let(:json)` makes it easier to understand the code for newcomers.

SuperTux88: Then I think a test for `#to_json` is better, like we have it for xml: `"produces correct XML"` -> `"produces correct JSON"`, instead of just parsing it without any checks.

> Also I think let(:json) makes it easier to understand the code for newcomers.

👍

app/schemas/federation_entities.json
cmrd-senya: I would prefer to keep the schema in a separate project. In theory the schema can be used in a different compatible implementation which is written in a different language and that has nothing to do with the ruby implementation of diaspora federation. Currently it is hardly possible that someone implements compatible software without reading our ruby source code, but in future it should be possible to implement protocol by using specifications and schemas and without touching ruby implementation.

But then the question raises: how should we deliver the schema to the `diaspora_federation` gem? We could create a gem for this, use git submodule in this repository or even download schema by its id URL. Using a gem is simple and robust, but using id URL to download schemas is better since it uses the "built-in" schema distribution approach (the fact that it is accessible via that URL) and doesn't need any extra distribution efforts. However if we choose the latter we'll need some caching method so that we don't have to download the schema on each rspec execution.

SuperTux88: > I would prefer to keep the schema in a separate project.

👍 

> But then the question raises: how should we deliver the schema to the `diaspora_federation` gem?

I think a gem is the best solution, since I don't want that every pod needs to download the schema every time. We can also copy the schema (like we did for the node-info schema in diaspora), but I think a gem that we can add as dependency is better than copy the schema every time there is a new version.

cmrd-senya: Let's decide on gem name then. My propositions:
`federation_entities.json_schema`
`federation_entities_json_schema`
`diaspora_federation_json_schema`

SuperTux88: Gem names don't contain a `.` (see: http://guides.rubygems.org/name-your-gem/), and also I think it should contain `diaspora`. So `diaspora_federation_json_schema` or maybe `diaspora_federation-json_schema`?

SuperTux88: And I thought about separating the schema again: So when we want to add a property to a entity, do we need to change the schema then? Or do we just keep the old schema since it can have additional properties anyway?

If we need to change the schema everytime we change an entity I still think we should keep the entities and the schema together, both the project and the version number, because otherwise it is too complicated to extend an entity, so you first need to extend the schema, then release it, then change the federation-library, release it and then add support for the property in diaspora. Then it's **much** easier to just add the property here in the federation library, both in the entity and the schema and then set the schema version to the next release-version of this gem, and release both together.

If we only change the schema from time to time and can add properties to entities without changing the schema, I'm fine with separate it to it's own project and own versioning.

cmrd-senya: They can safely contain a `.` (e.g. see [here](https://rubygems.org/gems/capistrano-db-tasks.senya)). `-` is normally used as ruby namespace separator, which is not strictly the case, but can be used. `diaspora_federation_json_schema` doesn't look very neat.

Also I'm not sure we need `diaspora` there, since we tend to avoid using `diaspora` in protocol related aspects (like we replaced `diaspora_handle` with `author` in entities and so on).

cmrd-senya: Okay, then I think that for now it is maybe better to keep them together. In future, when protocol stabilizes and we have a number of different implementations we'll have to keep and change them separately anyway. But since we still make many changes and develop the protocol intensively perhaps it's better to have them side by side here so it is easier for us to maintain. We can change this in future and move it out into a separate gem.

SuperTux88: Yes, I think that's the best solution for now (and we can always separate it later). Sorry for the back and forth here, but that just came to my mind that it probably makes the development much more complicated when we split it (but still need to update it every time).

And I also would keep the version together for now, so first release with this would be 0.2.0, and if then 0.2.1 maybe only is a bugfix release without entity changes, so I'd simply keep the schema version at 0.2.0, and if in 0.2.2 we add a property somewhere then I'd move the schema version to 0.2.2. I think that is OK to have gaps between schema versions here (maybe we can add a list with schema versions and schema changelog somewhere?).

cmrd-senya: Ok, then I'll set the version to 0.2.0 right away so you don't have to change it. We can also use `-dev` suffix between the releases for the case when between the realeases there are several commits touching the schema file. This way we won't have two different files in our commit history with the same non-dev version and will help to avoid issues when someone uses unreleased source code.

So I commit 0.2.0-dev now, you change it to 0.2.0 when you make a release. If in between 0.2.2 and 0.2.3 there are changes to the schema we change the version to 0.2.3-dev and then change it to 0.2.3 on release of 0.2.3. This way we avoid assigning normal version numbers (0.2.2 and 0.2.3) to intermidiate versions which may differ from both 0.2.2 and 0.2.3 releases.

> maybe we can add a list with schema versions and schema changelog somewhere?

I think we can create a page at https://diaspora.github.io/diaspora_federation.

SuperTux88: Sounds good to me. What do you think about also put the schemas at https://diaspora.github.io/diaspora_federation/schemas (at least for now while it is managed inside this gem)?

cmrd-senya: >  What do you think about also put the schemas at https://diaspora.github.io/diaspora_federation/schemas (at least for now while it is managed inside this gem)?

No objections.

cmrd-senya: So I think that we publish the schemas only for non-dev versions? So we'll have the
`"id": "https://diaspora.github.io/diaspora_federation/schemas/federation_entities-0.2.0-dev.json"` to identify the version in schema now, but we'll actually upload the schema to that address only when we make release and switch from `0.2.0-dev` to `0.2.0`? So as for the repository we'll have `app/schemas/federation_entities.json` file to track the last version and we'll have `docs/schemas` folder with all the released schema versions copied there.

SuperTux88: Yes 👍 

About `app/schemas`: Is that the correct place to put schemas? Also: In which gem should we package the schema? Should we add it to `diaspora_federation`? Or we can already create a separate gem for the schema `diaspora_federation-json_schema`, but release it together with everything else (like we already do for `-rails` and `-test`)?

cmrd-senya: As for documentation at  https://diaspora.github.io/diaspora_federation, we don't have any documentation there for the changes in this PR. Do you want me to add any documentation on our new JSON features within this PR?

> About app/schemas: Is that the correct place to put schemas?
We can move it to `lib/diaspora_federation/schemas` I think.

> Also: In which gem should we package the schema? Should we add it to diaspora_federation? Or we can already create a separate gem for the schema diaspora_federation-json_schema, but release it together with everything else (like we already do for -rails and -test)?

Hmm, why not, let's make a separate gem.

SuperTux88: > Do you want me to add any documentation on our new JSON features within this PR?

I wont block this PR for the documentation, but I'm still waiting for feedback from @denschub, so if you have the time to add it, do it in the PR. If everything else is finished in the PR, I'll merge it (so I can start with my refactorings), and you can add the documentation later.

> let's make a separate gem.

👍

cmrd-senya: Then adding a schema changelog page can be considered as a part of this documentation work, I think.

cmrd-senya:
> Merge branch 'develop' into to_json

Perhaps that was a mistake :)

cmrd-senya:
Anyway, I'll squash and rebase it as soon as the review process is complete.

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,366 @@
...
+        "entity_data": {
+          "type": "object",
+          "properties": {
+            "guid": { "type": "string" },
+            "public": { "type": "boolean" },
+            "created_at": { "type": "string" },
+            "provider_display_name" : { "type": "string" }
+          },
+          "required": [
+            "guid", "public", "created_at", "provider_display_name"
```
SuperTux88: `provider_display_name` is optional.

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,366 @@
...
+                  ]
+                },
+
+                "photos": {
+                  "type": ["array", "null"],
+                  "items": { "$ref": "#/definitions/photo" }
+                }
+              },
+
+              "required": ["text", "location", "poll", "photos"]
```
SuperTux88: `location`, `poll` and `photos` are also optional.

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,366 @@
...
+            "last_name": { "type": ["string", "null"] },
+            "gender": { "type": ["string", "null"] },
+            "bio": { "type": ["string", "null"] },
+            "birthday": { "type": ["string", "null"] },
+            "location": { "type": ["string", "null"] },
+            "image_url": { "type": ["string", "null"] },
+            "author": { "type": "string" }
+          },
+          "required": [
+            "first_name", "last_name", "gender", "bio", "birthday", "location", "image_url", "author"
```
SuperTux88: The only field required here is `author`, see https://diaspora.github.io/diaspora_federation/entities/profile.html

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,366 @@
...
+            "created_at": {"type": "string"},
+            "remote_photo_path": {"type": "string"},
+            "remote_photo_name": {"type": "string"},
+            "text": {"type": ["null", "string"]},
+            "status_message_guid": {"type": "string"},
+            "width": {"type": "number"},
+            "height": {"type": "number"}
+          },
+          "required": [
+            "guid", "remote_photo_path", "remote_photo_name", "status_message_guid", "width", "height"
```
SuperTux88: `author` is missing and required, `public` and `created_at` are also required, `status_message_guid` is optional.

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,366 @@
...
+    {"$ref": "#/definitions/photo"},
+    {"$ref": "#/definitions/poll"},
+    {"$ref": "#/definitions/poll_answer"}
+  ],
+
+  "definitions": {
+    "signature": {
+      "type": "string",
+      "minLength": 30
+    },
```
SuperTux88: Maybe also add a type for `guid` here, to be at least 16 and at most 255 chars long?

lib/diaspora_federation/parsers/json_parser.rb
```diff
@@ -0,0 +1,60 @@
+module DiasporaFederation
+  module Parsers
+    # This is a parser of JSON serialized object. JSON object format is defined by
+    # JSON schema which is available at [...].
+    # TODO: We must publish the schema at a real URL
```
SuperTux88: TODO :)

lib/diaspora_federation/parsers/relayable_json_parser.rb
```diff
@@ -0,0 +1,24 @@
+module DiasporaFederation
+  module Parsers
+    # This is a parser of JSON serialized object, that is normally used for parsing data of relayables.
+    # Assumed format differs from the usual entity by additional "property_order" property which is used to
+    # compute signatures deterministically.
+    # Input JSON for this parser is expected to match "/definitions/relayable" subschema of the JSON schema at [...].
```
SuperTux88: `[...]` :)

lib/diaspora_federation/parsers/json_parser.rb
cmrd-senya: Our schema URL contains a version number: `https://diaspora.github.io/diaspora_federation/schemas/federation_entities-0.2.0-dev.json`. If we write it here with the version number, then we'll have to update the link everytime we update the version number. Maybe we should make a symlink to the latest instead?

SuperTux88: We can add a `-latest` or simply `federation_entities.json` symlink for it.

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,373 @@
...
+            "created_at": {"type": "string"},
+            "remote_photo_path": {"type": "string"},
+            "remote_photo_name": {"type": "string"},
+            "text": {"type": ["null", "string"]},
+            "status_message_guid": {"$ref": "#/definitions/guid"},
+            "width": {"type": "number"},
+            "height": {"type": "number"}
+          },
+          "required": [
+            "guid", "remote_photo_path", "remote_photo_name", "width", "height", "author", "public", "created_at"
```
SuperTux88: Just a question: Should we order the `required` array somehow? Maybe same order as in `properties` above?

cmrd-senya: It makes no difference, but I'm okay with ordering them as in properties.

SuperTux88: Yes, I know that it technically doesn't make a difference, but I think it's better readable for humans :)

lib/diaspora_federation/schemas/federation_entities.json
```diff
@@ -0,0 +1,373 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://diaspora.github.io/diaspora_federation/schemas/federation_entities-0.2.0-dev.json",
```
denschub: (sorry I took so long, was busy with other stuff.)

My 👍 for using GitHub pages here. There is no real user impact if the site is unavailable and chances are low that we'll ever run into security issues on GitHub. Even if we do, we have local copies of the repos anyway. Also, updating the schemas is much easier on GitHub, since we just have to merge a PR. No reason to add a deploy chain to host it on diasporafoundation.org needed.

cmrd-senya:
So do I squash&rebase it now?

cmrd-senya:
Updated

SuperTux88:
Thanks 👍

For next time: Can you try to create multiple smaller commits with smaller steps (and split refactorings and adding new stuff), I think that makes it easier to understand the steps when needed afterwards from the git history :) So something like: "Extract xml parsing to parser-classes", "create json parsers", "Move SomeRelayable to spec/entities.rb", ...

cmrd-senya:
I'll try to move by smaller steps next time. It wasn't that obvious to me here, how to make steps smaller, but maybe that was to a lack of proper planning and designing.

SuperTux88:
The best way is to split refactorings from adding features, that would also already make the review easier, when I can see, that you only moved code from A to B, and in the next commit I see that you changed the same code to add a feature. And I think it's easier to create this commit-structure from the beginning (create a commit every time you finished a step), and then add the corrections as fixup to the corresponding commits. And if the we decide in the review to add more refactorings, then you can create a new commit for this refactoring.

PR #54, opened 2017-01-31
AccountMigration entity

cmrd-senya:
A new entity to use when diaspora ID of a person has changed for some reason (e.g. account migration, pod hostname change, etc)

I'm still in doubts about the name. `AccountRename` may be confusing. "Rename" is not a very accurate word for what we do: we change the ID, replace one with another. But `IDChange` was criticised for using "ID" word which is confusing because of different kinds of IDs being used in diaspora. Opinions?

### DIFF
--- a/docs/_entities/account_migration.md
+++ b/docs/_entities/account_migration.md
@@ -1,0 +1,59 @@
+---
+title: AccountMigration
+---
+
+This entity is sent when a person changes their diaspora* ID (e.g. when a user migration from one to another pod happens).
+
+## Properties
+
+| Property    | Type                         | Description                                                                          |
+| ----------- | ---------------------------- | ------------------------------------------------------------------------------------ |
+| `author`    | [diaspora\* ID][diaspora-id] | The diaspora\* ID of the closed account.                                             |
+| `person`    | [Profile][profile]           | New profile of a person                                                              |
+| `signature` | [Signature][signature]       | Signature that validates original and target diaspora* IDs with the new key of person |
+
+### Signature
+
+The signature base string is produced by concatenating the following substrings together, separated by semicolon (`:`):
+
+1) The entity name specifier: `AccountMigration`.
+
+2) diaspora\* ID of the closed account (old diaspora\* ID).
+
+3) diaspora\* ID of the replacement account (new diaspora\* ID).
+
+Example of a string:
+
+~~~
+AccountMigration:old-diaspora-id@example.org:new-diaspora-id@example.com
+~~~
+
+## Example
+
+~~~xml
+<account_migration>
+  <author>alice@example.org</author>
+  <profile>
+    <author>alice@newpod.example.net</author>
+    <first_name>my name</first_name>
+    <last_name/>
+    <image_url>/assets/user/default.png</image_url>
+    <image_url_medium>/assets/user/default.png</image_url_medium>
+    <image_url_small>/assets/user/default.png</image_url_small>
+    <birthday>1988-07-15</birthday>
+    <gender>Female</gender>
+    <bio>some text about me</bio>
+    <location>github</location>
+    <searchable>true</searchable>
+    <nsfw>false</nsfw>
+    <tag_string>#i #love #tags</tag_string>
+  </profile>
+  <signature>
+    07b1OIY6sTUQwV5pbpgFK0uz6W4cu+oQnlg410Q4uISUOdNOlBdYqhZJm62VFhgvzt4TZXfiJgoupFkRjP0BsaVaZuP2zKMNvO3ngWOeJRf2oRK4Ub5cEA/g7yijkRc+7y8r1iLJ31MFb1czyeCsLxw9Ol8SvAJddogGiLHDhjE=
+  </signature>
+</account_migration>
+~~~
+
+[diaspora-id]: {{ site.baseurl }}/federation/types.html#diaspora-id
+[profile]: {{ site.baseurl }}/entities/profile.html
+[signature]: {{ site.baseurl }}/federation/types.html#signature
--- a/lib/diaspora_federation/entities/relayable.rb
+++ b/lib/diaspora_federation/entities/relayable.rb
@@ -76,7 +76,8 @@ module DiasporaFederation
       end
 
       # Verifies the signatures (+author_signature+ and +parent_author_signature+ if needed).
-      # @raise [SignatureVerificationFailed] if the signature is not valid or no public key is found
+      # @raise [SignatureVerificationFailed] if the signature is not valid
+      # @raise [PublicKeyNotFound] if no public key is found
       def verify_signatures
         verify_signature(author, :author_signature)
 
--- a/lib/diaspora_federation/entities/relayable.rb
+++ b/lib/diaspora_federation/entities/relayable.rb
@@ -5,10 +5,7 @@ module DiasporaFederation
     # has a parent, identified by guid. Relayables are also signed and signing/verification
     # logic is embedded into Salmon XML processing code.
     module Relayable
-      include Logging
-
-      # Digest instance used for signing
-      DIGEST = OpenSSL::Digest::SHA256.new
+      include Signable
 
       # Order from the parsed xml for signature
       # @return [Array] order from xml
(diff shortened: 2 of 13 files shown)


SuperTux88:
> Opinions?

How about `AccountMigration`?

lib/diaspora_federation/entities/account_rename.rb
```diff
@@ -0,0 +1,25 @@
...
+      property :author, :string
+
+      # @!attribute [r] person
+      #   Holds new updated data of a person, including diaspora ID, public key, etc
+      #   @return [Person] person new data
+      entity :person, Entities::Person
+
+      # @return [String] string representation of this object
+      def to_s
+        "AccountRename:#{author}"
```
SuperTux88: I would add the new diaspora ID here too: `"AccountRename:#{author}:#{person.author}"`

spec/lib/diaspora_federation/entities/account_rename_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      <bio>#{data[:person].profile.bio}</bio>
+      <location>#{data[:person].profile.location}</location>
+      <searchable>#{data[:person].profile.searchable}</searchable>
+      <nsfw>#{data[:person].profile.nsfw}</nsfw>
+      <tag_string>#{data[:person].profile.tag_string}</tag_string>
+    </profile>
+    <exported_key>#{data[:person].exported_key}</exported_key>
+  </person>
+</account_rename>
+    XML
```
SuperTux88: Nit: Remove spaces before `XML` (yeah, I know, IDE ... I really need to figure out how to prevent rubymine from doing this ;) )

cmrd-senya: Can we make a rule for pronto/rubocop there?

SuperTux88: I don't know if/how this is possible, but if you want, you can try it :)

docs/_entities/account_migration.md
```diff
@@ -0,0 +1,51 @@
+---
+title: AccountMigration
+---
+
+This entity is sent when a person changes their diaspora ID (e.g. when a user migration from one to another pod
```
SuperTux88: Nit: Please write "diaspora" with \* in the documentation -> "diaspora\* ID"

cmrd-senya:
It is more or less finished, but #52 should be merged first, since I implemented `.from_hash` method for `AccountMigration` here so we don't have to change this later.

SuperTux88:
Looks "more or less" good so far ;) So waiting for #52 anyway.

Maybe we can extract some common signature-logic and the exceptions to a `SignedEntity` module?

docs/_entities/account_migration.md
```diff
@@ -0,0 +1,43 @@
...
+
+This entity is sent when a person changes their diaspora* ID (e.g. when a user migration from one to another pod happens).
+
+## Properties
+
+| Property    | Type                         | Description                                                                          |
+| ----------- | ---------------------------- | ------------------------------------------------------------------------------------ |
+| `author`    | [diaspora\* ID][diaspora-id] | The diaspora\* ID of the closed account.                                             |
+| `person`    | [Profile][profile]           | New profile of a person                                                              |
+| `signature` | [Signature][signature]       | Signature that validates original and target diaspora* IDs with the new key of person |
```
SuperTux88: Can you specify that the signature signs the String "AccountMigration:old-diaspora-id@example.org;new-diaspora-id@example.com". But probably not here in the table, I think a "Signature"-section below is better for that

cmrd-senya:
> Maybe we can extract some common signature-logic and the exceptions to a SignedEntity module?

Maybe just `Signable` is a better name for this module? It matches Relayable in a naming scheme then.

SuperTux88:
> Maybe just `Signable` is a better name for this module?

👍

cmrd-senya:
Updated. Please review.

docs/_entities/account_migration.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Signature
+
+The string that is being used to generate the signature or verify against it follows the pattern:
+
+~~~
+AccountMigration:#{old_diaspora_id};#{new_diaspora_id}
```
SuperTux88: This is ruby-syntax, so I'm not sure if we should write it that way in the documentation (because it should be independent from ruby), but also don't have a better idea right now.

For [Magic Signatures](https://diaspora.github.io/diaspora_federation/federation/magicsig.html#signature) we have it described with text. So maybe "`AccountMigration` and the old and the new diaspora\* ID, seperated with `:`"? Or maybe something more abstract like `AccountMigration:[old_diaspora_id]:[new_diaspora_id]`? But really, I don't know how to do this the best, what do you think?

(also it should be a `:` between the IDs, here and below at the example. And I just saw that I had it wrong at my comment above ... sorry)

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -5,10 +5,7 @@ module Entities
     # has a parent, identified by guid. Relayables are also signed and signing/verification
     # logic is embedded into Salmon XML processing code.
     module Relayable
-      include Logging
-
-      # Digest instance used for signing
-      DIGEST = OpenSSL::Digest::SHA256.new
+      include DiasporaFederation::Entities::Signable
```
SuperTux88: That can be only `include Signable`.

lib/diaspora_federation/entities/signable.rb
```diff
@@ -0,0 +1,53 @@
...
+        raise SignatureVerificationFailed, "wrong #{signature_key} for #{self}" unless valid
+
+        logger.info "event=verify_signature signature=#{signature_key} status=valid obj=#{self}"
+      end
+
+      # This method defines what data is used for a signature creation/verification
+      #
+      # @return [String] a string to sign
+      def signature_data
+        to_s
```
SuperTux88: This is the value needed for `AccountMigration`, but I'm not sure if we should add this as a default here. I think it's better to just raise a `NotImplementedError` here and need to implement it explicitly in every class that uses it.

docs/_entities/account_migration.md
cmrd-senya: Maybe it's okay to describe this in words. We have an example there, so we having a pattern is not necessary

SuperTux88: Yes, describe it in words + the example should do it 👍

cmrd-senya:
updated

SuperTux88:
Merged, thanks 👍

zotlabs:
I realise this is an old PR, but I've been going over it with an intent to work it into hubzilla or red while I'm still healthy enough to do so; and since the likelihood of anybody else doing this is slim.

Anyway I've been a bit confused (since this first floated by several months ago) with the text 

````
Signature that validates original and target diaspora* IDs with the new key of person
````

From my reading, this basically allows anybody to hijack a Diaspora account permanently by merely signing an AccountMigration message with the hijacker's key. 

If this is not the intent, this should probably be clarified. It seems much safer to sign with the old key; or better yet perhaps require participation of both keys in order to validate.

SuperTux88:
The whole message (the salmon) needs to be signed with the old key, and the `signature` needs to be signed with the new key. That way we can validate that the sender owns both keys/accounts.

zotlabs:
OK, got that. So salmon key_id is the original handle. I'm assuming either the old site or the new site can technically send it in this case and it should validate either way, but is it expected the new site will actually send it?

SuperTux88:
Yes, it doesn't matter who or what sends it, but in most cases the new site sends it (because the old site could be down, but with a backup of the key, the new site can still migrate the account).

zotlabs:
Thanks. All ready to go on the inbound side. The outbound side will be a bit more complicated as the default in red/hubzilla is a channel "copy" (useful for unplanned or brief outages, but incompatible with Diaspora). "Move" is kind of permanent. Will probably have to just ask folks what to do with their attached Diaspora connections whenever they create/edit/delete a clone.

zotlabs:
Oops - one more question. In the discussion the signed string was specified as 

AccountMigration:handle1;handle2 (semi-colon between the handles)

but in the commited spec it turned in AccountMigration:handle1:handle2 (colon between the handles).

Please advise.

SuperTux88:
The commited spec (colon) is correct.

PR #58, opened 2017-04-05
rspec without rails

SuperTux88:
I refactored the tests, so they don't need rails anymore. That way we can run the tests both, with and without rails, to ensure that everything also works without the rails dependencies.

I removed or replaced all test dependencies that had a dependency for rails and replaced activesupport methods used in the tests with own helper methods.

Locally everything seems to work, so lets see what travis says :)


spec/controllers/diaspora_federation/rails4_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        seed_url:      alice.url,
+        profile_url:   alice.profile_url,
+        atom_url:      alice.atom_url,
+        salmon_url:    alice.salmon_url,
+        subscribe_url: alice.subscribe_url,
+        guid:          alice.guid,
+        public_key:    alice.serialized_public_key
+      ).to_xml
+
+      get :legacy_webfinger, q: alice.diaspora_id
```
diaspora-code-review: Use keyword arguments instead of positional arguments for http call: `get`.

spec/controllers/diaspora_federation/rails4_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(response).to be_success
+      expect(response.body).to eq(webfinger_xrd)
+    end
+
+    it "404s when the person does not exist" do
+      get :legacy_webfinger, q: "me@mydiaspora.pod.com"
```
diaspora-code-review: Use keyword arguments instead of positional arguments for http call: `get`.

spec/controllers/diaspora_federation/rails4_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        photo_large_url:  alice.photo_default_url,
+        photo_medium_url: alice.photo_default_url,
+        photo_small_url:  alice.photo_default_url,
+        public_key:       alice.serialized_public_key,
+        searchable:       alice.searchable,
+        first_name:       alice.first_name,
+        last_name:        alice.last_name
+      ).to_html
+
+      get :hcard, guid: alice.guid
```
diaspora-code-review: Use keyword arguments instead of positional arguments for http call: `get`.

spec/controllers/diaspora_federation/rails4_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(response).to be_success
+      expect(response.body).to eq(hcard_html)
+    end
+
+    it "404s when the person does not exist" do
+      get :hcard, guid: "unknown_guid"
```
diaspora-code-review: Use keyword arguments instead of positional arguments for http call: `get`.

spec/controllers/diaspora_federation/rails4_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  describe ReceiveController, type: :controller, rails: 4 do
+    routes { DiasporaFederation::Engine.routes }
+
+    describe "POST #public" do
+      it "returns a 202 if queued correctly" do
+        expect_callback(:queue_public_receive, "<diaspora/>", true)
+
+        post :public, xml: "<diaspora/>"
```
diaspora-code-review: Use keyword arguments instead of positional arguments for http call: `post`.

spec/controllers/diaspora_federation/rails4_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        expect(response.code).to eq("202")
+      end
+    end
+
+    describe "POST #private" do
+      it "returns a 202 if the callback returned true" do
+        expect_callback(:queue_private_receive, "any-guid", "<diaspora/>", true).and_return(true)
+
+        post :private, guid: "any-guid", xml: "<diaspora/>"
```
diaspora-code-review: Use keyword arguments instead of positional arguments for http call: `post`.

spec/support/shared_validator_specs.rb
```diff
@@ -135,7 +135,7 @@ def alphanumeric_string(length)
   end
 
   it "must not be an arbitrary string or other object" do
-    ["asdf", Time.zone.today, 1234].each do |val|
+    ["asdf", Date.today, 1234].each do |val|
```
diaspora-code-review: Do not use `Date.today` without zone. Use `Time.zone.today` instead.

SuperTux88:
Hmm, I changed the `.rubocop.yml`, so there shouldn't be any errors ... but @diaspora-code-review seems to ignore the config from the PR?

SuperTux88:
And travis is green 🎉

spec/support/helper_methods.rb
```diff
@@ -0,0 +1,42 @@
...
+def change_time(time, options={})
+  new_hour  = options.fetch(:hour, time.hour)
+  new_min   = options.fetch(:min, options[:hour] ? 0 : time.min)
+  new_sec   = options.fetch(:sec, options[:hour] || options[:min] ? 0 : time.sec)
+
+  ::Time.utc(time.year, time.month, time.day, new_hour, new_min, new_sec)
+end
+
+# indent helper
+def indent(string, amount)
```
cmrd-senya: Don't you think it might be better to implement this method the same way as ActiveSupport does, by reopening `String` class?
So we have here
```ruby
class String
  def indent(amount)
    gsub(/^/, " " * amount)
  end
end
```

The same question goes for `change_time` above.

SuperTux88: I thought about that, but that would allow to also use this methods in the library-code without noticing, and when doing it with the helper methods, that isn't the case. And what I wanted to achieve with this is that travis fails when we accidentally use stuff that is only available in activesupport.

SuperTux88:
> So we only require actionpack now in order to have ActionController::Base available?

Yes, so diaspora_federation-rails doesn't need all rails dependencies anymore (but in most apps this doesn't make a difference)

> Does Rails related tests run when I launch bundle exec rspec now?

By default it uses the main Gemfile, that includes rails, so it also runs the rails related tests (controllers).

> If we have Rails dependency in the Gemfile wouldn't it be possible to accidentaly use something from Rails in the rspec tests again?

By default you wouldn't detect if you use something from rails, but on travis this will fail. You can also run rspec with a different Gemfile with `BUNDLE_GEMFILE=test/gemfiles/no-rails.Gemfile rake` locally. If you change something at the main Gemfile, I added a rake task to update/generate those test gemfiles (`rake gemfiles:generate`).

(This comment is written from my smartphone, I hope the commands work, because I haven't tested them now ;) )

SuperTux88:
Merged, thanks for the reviews @cmrd-senya and @denschub ❤️

SuperTux88:
> If we have Rails dependency in the Gemfile wouldn't it be possible to accidentaly use something from Rails in the rspec tests again?

@cmrd-senya: I now understand what you meant with that (sorry, I misunderstood that the first time), and added a check in the travis.sh, that checks if the no-rails.Gemfile.lock really doesn't contain any rails dependencies. So we detect if we accidentaly add something that depends on rails in rspec tests again. See: https://github.com/diaspora/diaspora_federation/commit/f08db93d74c87e1fb0e399232d88fccbf4e3dc47

PR #59, opened 2017-04-23
Add public flag to profile entity

SuperTux88:
Diaspora has this flag for some time now, but it is pod-local only. With this change we can also federate this information. We can also federate public profiles publicly, so public profiles with many contacts need to be federated only once (would be useful for the diaspora\* HQ account for example).


SuperTux88:
Thanks for the review @denschub and @cmrd-senya 👍

Flaburgan:
Thank you very much for this change, I'm always sad when a feature is available only locally.

Issue #39, opened 2016-08-06
Public-key and GUID moved from WebFinger to hCard


SuperTux88:
I already mentioned this here: https://github.com/diaspora/diaspora/issues/6405#issuecomment-141598962, but I discovered, that there is no issue for this in this repository.

The public-key and the GUID were moved from the WebFinger-xml to the the hCard: 
- `rel="http://joindiaspora.com/guid"` is now deprecated in the webfinger-xml, the guid is now searched in the HCard with the selector `.uid`
- `rel="diaspora-public-key"` is also deprecated in the webfinger-xml, the public key is now searched in the HCard with the selector `.key`

Diaspora supports this change since nearly a year now (see diaspora/diaspora#6310) and the old method is still supported for backward-compatibility, but the backward-compatibility will be removed soon.

PR #60, opened 2017-04-24
Cleanup webfinger deprecations

SuperTux88:
This removes backward-compatibility for the changes from #39 and breaks with pod older than 0.5.3.0.


SuperTux88:
Thanks for the review @denschub 🍪

PR #61, opened 2017-04-27
Improve WebFinger generation

SuperTux88:
This makes the WebFinger generation more flexible, so it's better usable for other projects.

* It makes not needed properties optional for generation (was already optional for parsing).
* It allows to add additional elements (unrelated to the diaspora protocol).
  Diaspora also needs this for OpenIDConnect when we switch to RFC7033 WebFinger (which will be the next step).

### DIFF
--- a/lib/diaspora_federation/discovery/web_finger.rb
+++ b/lib/diaspora_federation/discovery/web_finger.rb
@@ -38,7 +38,7 @@ module DiasporaFederation
       # @!attribute [r] alias_url
       #   @note could be nil
       #   @return [String] link to the users profile
-      property :alias_url, :string
+      property :alias_url, :string, default: nil
 
       # @!attribute [r] hcard_url
       #   @return [String] link to the +hCard+
@@ -50,7 +50,7 @@ module DiasporaFederation
 
       # @!attribute [r] profile_url
       #   @return [String] link to the users profile
-      property :profile_url, :string
+      property :profile_url, :string, default: nil
 
       # @!attribute [r] atom_url
       #   This atom feed is an Activity Stream of the user's public posts. diaspora*
@@ -61,18 +61,18 @@ module DiasporaFederation
       #   Note that this feed MAY also be made available through the PubSubHubbub
       #   mechanism by supplying a <link rel="hub"> in the atom feed itself.
       #   @return [String] atom feed url
-      property :atom_url, :string
+      property :atom_url, :string, default: nil
 
       # @!attribute [r] salmon_url
       #   @note could be nil
       #   @return [String] salmon endpoint url
       #   @see https://cdn.rawgit.com/salmon-protocol/salmon-protocol/master/draft-panzer-salmon-00.html#SMLR
       #     Panzer draft for Salmon, paragraph 3.3
-      property :salmon_url, :string
+      property :salmon_url, :string, default: nil
 
       # @!attribute [r] subscribe_url
       #   This url is used to find another user on the home-pod of the user in the webfinger.
-      property :subscribe_url, :string
+      property :subscribe_url, :string, default: nil
 
       # +hcard_url+ link relation
       REL_HCARD = "http://microformats.org/profile/hcard".freeze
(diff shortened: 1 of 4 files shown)


lib/diaspora_federation/discovery/web_finger.rb
```diff
@@ -146,15 +165,38 @@ def to_s
         end
       end
 
+      def add_aliases_to(doc)
```
cmrd-senya: Maybe we should use `doc` as a method instead of passing it as a parameter?
```ruby
def doc
  @doc ||= XrdDocument.new
end
```

It doesn't make much difference, but using a method instead of parameter passing feels like a better design.

spec/lib/diaspora_federation/discovery/web_finger_spec.rb
```diff
@@ -40,6 +49,52 @@ module DiasporaFederation
         wf = Discovery::WebFinger.new(data)
         expect(wf.to_xml).to eq(xml)
       end
+
+      it "creates minimal XML document" do
+        wf = Discovery::WebFinger.new(
+          acct_uri:  "acct:#{person.diaspora_id}",
```
cmrd-senya: We have `acct` shortcut for this string, see line 4.

spec/lib/diaspora_federation/discovery/web_finger_spec.rb
```diff
@@ -40,6 +49,52 @@ module DiasporaFederation
...
+  <Link rel="http://microformats.org/profile/hcard" type="text/html" href="#{person.hcard_url}"/>
+  <Link rel="http://joindiaspora.com/seed_location" type="text/html" href="#{person.url}"/>
+  <Link rel="http://portablecontacts.net/spec/1.0" href="https://pod.example.tld/poco/trouble"/>
+  <Link rel="http://webfinger.net/rel/avatar" type="image/jpeg" href="http://localhost:3000/assets/user/default.png"/>
+</XRD>
+XML
+
+        wf = Discovery::WebFinger.new(
+          {
+            acct_uri:  "acct:#{person.diaspora_id}",
```
cmrd-senya: ditto

lib/diaspora_federation/discovery/web_finger.rb
```diff
@@ -38,7 +38,7 @@ class WebFinger < Entity
       # @!attribute [r] alias_url
       #   @note could be nil
       #   @return [String] link to the users profile
-      property :alias_url, :string
+      property :alias_url, :string, default: nil
```
cmrd-senya: Maybe we don't need to define a special `alias_url` property anymore and can rely on `additional_data[:aliases]` as the only method of defining aliases in the WebFinger (Xrd document)?

lib/diaspora_federation/discovery/web_finger.rb
```diff
@@ -146,15 +165,38 @@ def to_s
         end
       end
 
+      def add_aliases_to(doc)
+        doc.aliases << alias_url if alias_url
+
+        return unless additional_data[:aliases]
+
+        additional_data[:aliases].each do |a|
```
cmrd-senya: Or maybe even just
```ruby
doc.aliases.concat(additional_data[:aliases])
```
?

lib/diaspora_federation/discovery/web_finger.rb
SuperTux88: I don't think we should do this, because as you wrote, it doesn't make much difference, but it breaks for example when you call `to_xml` twice on the same object.

lib/diaspora_federation/discovery/web_finger.rb
SuperTux88: We need the `if` here, because `concat` doesn't work with `nil`.

SuperTux88:
Thanks for the review 👍

I edited the second commit with your comments, so please have a look again.

cmrd-senya:
Code looks good to me :+1: 

It looks like there is some issue with travis unrelated to your PR

SuperTux88:
Yes, travis seams to have some issues, see: https://github.com/travis-ci/travis-ci/issues/7685

SuperTux88:
Travis fixed :)

SuperTux88:
Thanks for the reviews @cmrd-senya and @denschub ❤️

Issue #36, opened 2016-06-17
Message Entity non-relayable


SuperTux88:
Private messages are currently always relayed over the owner of the conversation. Therefore they need an additional signature in the xml. This can be simplified, if the author of the message itself sends the message to all participants, because all participants are known by every participant. This means, that the fields `author_signature` and `parent_author_signature` aren't needed anymore. Also the `Message` entity have a duplicate field `parent_guid` which contains the same information as in `conversation_guid`, so `parent_guid` can be dropped in the future.
### v0.1.0 (for diaspora 0.6.0.0)
- [x] send signed messages to all participants
- [x] relay to all participants (from conversation-parent)
- [x] don't check signatures if received from author
- [x] check `author_signature` if received from conversation-author
### v0.2.0 (for diaspora 0.7.0.0)
- [x] send message without signatures and without `parent_guid`
- [x] don't relay it to participants
- [x] receive only from author

Issue #35, opened 2016-06-17
Participation Entity non-relayable


SuperTux88:
Participations are currently relayed like comments and likes to every pod. This is not necessary, because the information is only important for the Pod that owns the parent object, because he needs to send updates to every participant. On every other pod, it only makes the Database big. So we can change this to a non-relayable, which mean, that the fields `author_signature` and `parent_author_signature` aren't needed anymore.
### v0.1.0 (for diaspora 0.6.0.0)
- [x] send signed participations to the parent-author-pod
- [x] receive only from author (drop relayed participations on receive)
- [x] don't check signatures on receive
### v0.2.0 (3-4 diaspora minor-releases after diaspora 0.6.0.0)
- [ ] send participations without signatures

Issue #32, opened 2016-03-23
new entity for "start sharing with"


SuperTux88:
I discussed this with @annando a few days ago. The main reason was, that the "stop sharing with you" currently is a `retraction` with the type `Person` which is weird, because the retraction should always be the same type as the original message. And we thought that `Request` is a weird name anyway for "start sharing with" and friendica has `following` and `sharing`.
### new `Contact` entity

I created a new entity for this with a better name for this action. The xml that diaspora would send look like this:

```
<contact>
  <author>alice@example.com</author>
  <recipient>bob@example.org</recipient>
  <following>true</following>
  <sharing>true</sharing>
</contact>
```

Diaspora doesn't support `following` and `sharing` yet, so both would be true now, but this change sometime: https://discourse.diasporafoundation.org/t/sharing-vs-following/524

If someone stops sharing/following, he sends the same message with false for `following`/`sharing`.
### v0.1.0 (for diaspora 0.6.0.0)
- [x] send the old Request entity
- [x] send the old retraction with type `Person`
- [x] can receive and parse old Request entity
- [x] can receive and parse new Contact entity
- [x] can receive and parse old retraction with type `Person`
### v0.2.0 (for diaspora 0.7.0.0)
- [x] send the new Contact entity for start and stop sharing
### v0.3.0 (probably for diaspora 0.8.0.0)
- [ ] remove old Request entity and cleanup.

annando:
Do you think "author" is necessary? This is a message type that surely isn't relayed. So we have the author already through the salmon part.

To clear the direction of "following" and "sharing" in this context:
* "following": "author" wants to read the content of "recipient"
* "sharing": "author" wants to send content to "recipient"

Correct?

BTW: We don't need the retraction here. We simply could set one or both of these values to "false" to signalise the "unsharing" or "unfollowing". A simple retraction without specifying the direction would be too unspecified.

SuperTux88:
I think we should have the `author` as we have it for all other entities too.

following/sharing: correct.

I think a retraction is better, because it is more what also happens in the database. If someone switches from `following` to `both`, we send an updated `contact` entity and update the state in the database. But if someone disconnects completely we send a `retraction` and delete the contact in the database.

redmatrix:
[...]

Issue #30, opened 2016-02-15
Salmon Refactoring


SuperTux88:
We use currently a `<diaspora>` XML wrapper for the salmon magic envelope. The main reason for this wrapper is to send an `author_id` in the header. But we can use the `key_id` from the magic envelope for that. The second reason is for the aes-crypt for private-messages. But because we encrypt the magic envelope payload and the header separately, we need two aes keys.
## public messages

For public messages we can simply use the normal magic envelope as specified here: https://cdn.rawgit.com/salmon-protocol/salmon-protocol/master/draft-panzer-magicsig-01.html
- Add `key_id` with the base64url encoded diaspora-id (see 3.1)
- No additional wrapping with own xml, no diaspora-header.
- send it with `Content-Type: application/magic-envelope+xml` as request-body (no `xml=` param) (see [salmon](https://cdn.rawgit.com/salmon-protocol/salmon-protocol/master/draft-panzer-salmon-00.html#RPF))
- the data-payload of the magic-env would stay the same (changes from #28 + #29)

This is much simpler and it is basically the salmon/magic-env spec.
## private messages

For private messages we need an encryption again. But instead of using the complicated "EncryptedSlap" with separately encrypted header and magic-env payload (another hack in the magic-env to encrypt/decrypt the data-payload). we can simply encrypt the whole magic-env (so the magic-env can be handled exactly the same as for public-messages, no additional hacks needed). For the crypt-wrapper I used a similar json-structure as for the encrypted_header from the "EncryptedSlap":

```
{
  "aes_key": "...",
  "encrypted_magic_envelope": "..."
}
```

The `aes_key` is the same as for the `encrypted_header`: It contains the base64 encoded AES `key` and `iv` as json, encrypted with the recipients public key and base64 encoded again:

```
{
  "key": "...",
  "iv": "..."
}
```

The `encrypted_magic_envelope` is the magic-env encrypted with the `aes_key` and base64 encoded. After decoding and decrypting it can be parsed like a normal magic envelope(see public messages). It also contains the `key_id`.

This crypt-json should be transmitted also directly in the request-body as `Content-Type: application/json`.

Because the new messages don't use the old unnecessary `xml=` parameter, we can decide with the content-type and the param, if we received a new or an old message for the transition period.
### v0.1.0 (for diaspora 0.6.0.0)
- [x] send the old Slap and EncryptedSlap (`<diaspora>...</diaspora>`)
- [x] send the `key_id` in the magic envelope
[...]

PR #62, opened 2017-05-02
Next step for the new protocol

SuperTux88:
This contains all changes for the second step of the new protocol (sending new protocol, but still be able to receive both versions).

* #26: Remove legacy signature order and order by property order in entity
* #27: Remove sign-code and prevent creation of `SignedRetraction` and `RelayableRetraction`
* #28: Don't wrap entity-XML in `<XML><post>...</post></XML>` for sending
* #29: Send messages with new property names
* #30: Send the raw magic envelope and new encrypted magic envelope with crypt-json-wrapper
* #32: Prevent creation of `Request`
* #35: Make `Participation` entity non-relayable
* #36: Make `Message` entity non-relayable

I also pushed the changes needed for diaspora here: diaspora/diaspora#7436

I tested these changes between different versions (0.6.0.0, latest-develop and develop with these changes), and everything seams to work 🎉. This doesn't contain the fix for diaspora/diaspora#4269, so we don't break with pods older than 0.6.3.0 yet (can be changed in a later PR). @annando still needs to test if all these changes work with friendica too.


lib/diaspora_federation/entities/comment.rb
```diff
@@ -21,6 +17,12 @@ class Comment < Entity
       #   Comment entity creation time
       #   @return [Time] creation time
       property :created_at, :timestamp, default: -> { Time.now.utc }
+
+      # Remove "created_at" when no order was received.
+      # @deprecated TODO: Remove this, this will break compatibility with pods older than 0.6.3.0.
```
SuperTux88: The can handle it, but they would generate the wrong formatted timestamp when relaying (That was fixed in 0.6.3.0), and that would break relaying for posts on old pods.

lib/diaspora_federation/entities/relayable.rb
```diff
@@ -103,6 +99,17 @@ def to_json
         }
       end
 
+      # The order for signing
+      # @return [Array]
+      def signature_order
+        if @signature_order
```
SuperTux88: This doesn't work, because the entities are frozen after creation, but I used `||` instead of `||=` for now. I think I will simply always set `@signature_order` in the constructor (so set the default order there too) after the hack for comments is gone, but for now I think that is OK.

spec/lib/diaspora_federation/entities/relayable_spec.rb
```diff
@@ -26,8 +26,8 @@ module DiasporaFederation
 
     describe "#verify_signatures" do
       it "doesn't raise anything if correct signatures with legacy-string were passed" do
```
SuperTux88: You're right, and this test didn't even do what it's description said ;) But since we don't differentiate new or old in `#verify_signature` (we just need to respect the order), I changed the tests to test that, and removed the `new signatures` part, because it was obsolete now.

SuperTux88:
@cmrd-senya thanks for the review, I corrected your suggestions 👍 waiting for the second part of the review :)

spec/integration/comment_integration_spec.rb
```diff
@@ -191,21 +201,21 @@ module DiasporaFederation
       end
 
       it "relays legacy signatures and xml" do
```
SuperTux88: You're right, the focus is on "order" and not "signature", so I changed that. I keeped "legacy", so it's "legacy order" and "new order" for now (instead of "random order 1 or 2"). I also did some other tweaks to make `let`s and test descriptions better understandable. I think that's OK for now, since these tests will change anyway, when we drop legacy xml format parsing. And I think we can drop the different order tests then, because the "new data" tests also includes that somehow.

SuperTux88:
Thanks for the reviews @cmrd-senya and @denschub 🍪

PR #63, opened 2017-05-14
Implement RFC 7033 Webfinger

SuperTux88:
Needed for diaspora/diaspora#5963

### DIFF
--- a/docs/discovery/webfinger.md
+++ b/docs/discovery/webfinger.md
@@ -6,7 +6,7 @@ diaspora\* uses an old draft of [WebFinger][webfinger-draft] to discover users f
 
 {% include warning_box.html
    title="Old WebFinger"
-   content="<p>diaspora* doesn't yet support the RFC 7033 WebFinger!</p>"
+   content="<p>diaspora* doesn't yet fully support the RFC 7033 WebFinger!</p>"
 %}
 
 ## WebFinger endpoint discovery
@@ -54,7 +54,7 @@ Content-Type: application/xrd+xml; charset=utf-8
 ~~~xml
 <?xml version="1.0" encoding="UTF-8"?>
 <XRD xmlns="http://docs.oasis-open.org/ns/xri/xrd-1.0">
-  <Link rel="lrdd" type="application/xrd+xml" template="https://example.org/webfinger?q={uri}"/>
+  <Link rel="lrdd" type="application/xrd+xml" template="https://example.org/.well-known/webfinger.xml?resource={uri}"/>
 </XRD>
 ~~~
 
@@ -70,7 +70,7 @@ of the searched person.
 #### Example
 
 ~~~
-GET /webfinger?q=acct:alice@example.org
+GET /.well-known/webfinger.xml?resource=acct:alice@example.org
 Host: example.org
 ~~~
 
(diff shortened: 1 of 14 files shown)


SuperTux88:
Thanks for your feedback, I changed the spec.

docs/discovery/webfinger.md
```diff
@@ -6,7 +6,7 @@ diaspora\* uses an old draft of [WebFinger][webfinger-draft] to discover users f
 
 {% include warning_box.html
    title="Old WebFinger"
-   content="<p>diaspora* doesn't yet support the RFC 7033 WebFinger!</p>"
+   content="<p>diaspora* doesn't yet fully support the RFC 7033 WebFinger!</p>"
```
cmrd-senya: There is a line above in the doc that says `diaspora* uses an old draft of WebFinger to discover users from other pods.`. This is no longer true. We can change that to `diaspora* supports RFC 7033 WebFinger to discover users from other pods.`. Here we have this "Old WebFinger" notice. If we have the above text fixed, then we can make more sense to change the context of the box to something like `diaspora* below version 0.7.0.0 supports only Old draft WebFinger. RFC 7033 WebFinger is supported from 0.7.0.0 and higher. This document focuses on old WebFinger.`.

Also we have "Additional information and specifications" box at the page footer. We could actually add RFC 7033 WebFinger specification link there.

spec/controllers/diaspora_federation/webfinger_controller_spec.rb
```diff
@@ -35,31 +35,77 @@ module DiasporaFederation
...
     end
 
-    describe "GET #legacy_webfinger", rails: 5 do
-      it "succeeds when the person exists" do
-        get :legacy_webfinger, params: {q: alice.diaspora_id}
+    describe "GET #webfinger", rails: 5 do
+      it "uses the JRD format as default" do
+        get :webfinger, params: {resource: alice.diaspora_id}
         expect(response).to be_success
+        expect(response.header["Content-Type"]).to include "application/jrd+json"
```
cmrd-senya: Maybe add a check for CORS header here as well or with another example?

docs/discovery/webfinger.md
SuperTux88: > `diaspora* uses an old draft of WebFinger to discover users from other pods.`

This is still true, because diaspora still uses the old WebFinger for discovery, this PR adds support that we provide the new RFC 7033, so we can use it later (and other software can use it).

> Also we have "Additional information and specifications" box at the page footer. We could actually add RFC 7033 WebFinger specification link there.

👍

spec/controllers/diaspora_federation/webfinger_controller_spec.rb
```diff
@@ -35,31 +35,77 @@ module DiasporaFederation
...
+
+        it "returns a application/jrd+json" do
+          get :webfinger, format: :json, params: {resource: "acct:#{alice.diaspora_id}"}
+          expect(response.header["Content-Type"]).to include "application/jrd+json"
+        end
+
+        it "404s when the person does not exist" do
+          get :webfinger, format: :json, params: {resource: "me@mydiaspora.pod.com"}
+          expect(response).to be_not_found
+        end
```
cmrd-senya: RFC WebFinger requires that if we miss the `resource` parameter we fail with `400` status. It is how we work now, but since this is a specification, maybe we should add an example for this case as well?

SuperTux88:
I updated the PR, I can rebase/squash it when it's OK for you.

spec/controllers/diaspora_federation/webfinger_controller_spec.rb
SuperTux88: I wasn't able to actually test the `400` (everything I tried didn't work, it always raised the exception :( ) So I now added a test that checks that exception (which results in a `400` in production).

SuperTux88:
I squashed the commits.

cmrd-senya:
Merged as ea00874c939400e74ba88a6b586b903386a787a3, 12cd5b0090afe4eaf50fe7f7c9b1e60305a829bd, 6e51ae536b8dded8a7d2b977cb7c2861879b14db. 

Thanks :tada: !

Issue #64, opened 2017-05-14
Remove "parent_author_signature" from relayables


SuperTux88:
@jaywink asked a few weeks ago, why he needs to add 3 signatures when he sends a comment to his own post (`author_signature`, `parent_author_signature` and the signature in the Magic Envelope, all with the same key). So we can simplify how signatures for relayables are handled:

We can remove `parent_author_signature` completely and use the signature from the Magic Envelope instead, since we receive a relayable always via the parent author (except when we are the parent author). So instead of verifying the `parent_author_signature` we need to verify that the sender of the MagicEnv is the parent author.

There is a little edge case, when the author and the parent author is on the same pod (but not the same person), who sends the MagicEnv then? But I checked the diaspora implementation and also asked @annando for friendica, we already use the parent author in this case, so everything is fine.

There is a special case, when the author is also the parent author. In this case we don't even need to check the `author_signature`, the signature in the Magic Envelope is enough for that.

### v0.2.0 (for diaspora 0.7.0.0)
- [x] don't verify `parent_author_signature` anymore
- [x] verify that the sender of the Magic Envelope is the parent author or the parent is local and we receive the entity from the author directly.
- [x] don't verify `author_signature` if the author is also the parent author
### v0.3.0 (probably for diaspora 0.8.0.0)
- [ ] don't send `parent_author_signature` anymore
- [ ] don't send `author_signature` if the author is also the parent author

annando:
Friendica currently checks the ```parent_author_signature``` and will reject a comment or like when it is missing.

So it seems that for the upcoming 3.5.2 we already should remove the check.

SuperTux88:
Diaspora keeps sending the `parent_author_signature`, so no need to hurry :) 0.7.0.0 well stop checking the signature, and some future unplanned release (maybe 0.8.0.0) then stops sending the `parent_author_signature`, because current diaspora installation would also reject a comment or like when it is missing.

jaywink:
Thanks for taking note of the comment, this is a good improvement imho to simplify implementation 👍

PR #65, opened 2017-05-14
Refactor relayable signatures

SuperTux88:
Step 1 for #64

### DIFF
--- a/docs/_entities/comment.md
+++ b/docs/_entities/comment.md
@@ -16,7 +16,6 @@ See also: [Relayable][relayable]
 | `text`                    | [Markdown][markdown] (65535) | The comment text.                               |
 | `created_at`              | [Timestamp][timestamp]       | The create timestamp of the comment.            |
 | `author_signature`        | [Signature][signature]       | The signature from the author of the comment.   |
-| `parent_author_signature` | [Signature][signature]       | The signature from the parent entity author.    |
 
 ## Optional Properties
 
@@ -26,8 +25,6 @@ See also: [Relayable][relayable]
 
 ## Examples
 
-### From author
-
 ~~~xml
 <comment>
   <author>alice@example.org</author>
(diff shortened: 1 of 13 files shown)


cmrd-senya:
We probably should mark `parent_author_signature` as deprecated in our API reference:
```diff
diff --git a/lib/diaspora_federation/entities/relayable.rb b/lib/diaspora_federation/entities/relayable.rb
index 7a06ddf..21d765b 100644
--- a/lib/diaspora_federation/entities/relayable.rb
+++ b/lib/diaspora_federation/entities/relayable.rb
@@ -45,6 +45,7 @@ module DiasporaFederation
       #   This signature is required only when federating from upstream (parent) post author to
       #   downstream subscribers. This is the case when the parent author has to resend a relayable
       #   received from one of their subscribers to all others.
+      #   @deprecated
       #   @return [String] parent author signature
       #
       # @!attribute [r] parent
```

SuperTux88:
Thanks for the review.

About the `parent_author_signature` in JSON: I removed it completely, so it isn't even added to the JSON when generating it. And since it was never there in JSON (because JSON is new), we don't need it in the schema either. And it doesn't make sense to add it for exporting anyway, because when the exporting user is the parent, we don't need it, and when the exporting user isn't the parent, the `parent_author_signature` can't be included anyway.

I also fixed the other comments, so please have another look at it :)

docs/_entities/comment.md
```diff
@@ -50,7 +48,6 @@ See also: [Relayable][relayable]
   <parent_guid>c3893bf029e7013487753131731751e9</parent_guid>
   <text>this is a very informative comment</text>
   <author_signature>cGIsxB5hU/94+rmgIg/Z+OUvXVYcY/kMOvc267ybpk1pT44P1JiWfnI26F1Mta62UjzIW/SjeAO0RIsJRguaISLpXX/d5DJCMpePAZaZiagUbdgH/w4L++fXiPxBKkSm+PB4txxmHGN8FHjwEUJFHJ1m3VfU4w2JC8+IBU93eag=</author_signature>
-  <parent_author_signature>uzjxUSqR8DQBSBa6abY7R/s9DVzT6UAgTctRcUu5rV5o0iXJD2MR6kp6bsVH3nMbbNvOjwAtrdfz3SVHT2gD8M5PmoFagxK7m5T1c9FB0i+wknoAah0Si0c2sP/BPLnkQ83DgLjF+JZCzMX4sWKvYfyaMjnih1MtQILGyuiwA0E=</parent_author_signature>
```
SuperTux88: You are right, thanks for the hint. I removed the separation between "From author" and "From parent author" and squashed it to the commit where I already removed `parent_author_signature` from the documentation.

PR #67, opened 2017-06-05
Add created_at to comment

SuperTux88:
Since my [post on diaspora](https://nerdpol.ch/posts/635fb840f46f0134b8f952540061b601) some more podmins updated, so I think we can think about this again. I use the numbers from [the-federation.info](https://the-federation.info/diaspora), but subtracted the duplicates (librenet and datataffel).

We have currently 14970 active users last month and 14599 are on a pod with 0.6.3.0 or newer. That are 97.5%. That are 4% more than 2 Months ago. Until diaspora/diaspora#7436 gets merged to `develop` it'll still take some weeks (since we are still waiting for a friendica release and then + 2-3 weeks), so maybe even more podmins updates until then. (we can create another post or ask some podmins)

So 97.5% is OK for me, I don't think we need to wait longer. Any vetos? @denschub @svbergerem @cmrd-senya @Flaburgan 

(I don't think we need a bigger discussion/vote for this, but if anybody thinks we need it, just say it. But since we need this PR anyway sooner or later, I just created it :) )

This fixes diaspora/diaspora#4269

### DIFF
--- a/lib/diaspora_federation/entities/comment.rb
+++ b/lib/diaspora_federation/entities/comment.rb
@@ -17,12 +17,6 @@ module DiasporaFederation
       #   Comment entity creation time
       #   @return [Time] creation time
       property :created_at, :timestamp, default: -> { Time.now.utc }
-
-      # Remove "created_at" when no order was received.
-      # @deprecated TODO: Remove this, this will break compatibility with pods older than 0.6.3.0.
-      def signature_order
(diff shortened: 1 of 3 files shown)


Flaburgan:
Let's move ahead!

cmrd-senya:
Merged, thanks! :+1:

PR #70, opened 2017-07-21
Add DiasporaFederation::Schemas to access JSON schema

SuperTux88:
See https://github.com/diaspora/diaspora/pull/6726#discussion_r128658844

### DIFF
--- a/diaspora_federation.gemspec
+++ b/diaspora_federation.gemspec
@@ -17,7 +17,8 @@ Gem::Specification.new do |s|
   s.license     = "AGPL-3.0"
 
   s.files       = Dir["lib/**/*", "LICENSE", "README.md", "Changelog.md"] -
-                    Dir["lib/diaspora_federation/{engine,rails,test}.rb",
+                    Dir["lib/diaspora_federation/{engine,rails,schemas,test}.rb",
+                        "lib/diaspora_federation/schemas/*",
                         "lib/diaspora_federation/test/*",
                         "lib/tasks/*.rake"]
 
--- a/lib/diaspora_federation/schemas.rb
+++ b/lib/diaspora_federation/schemas.rb
@@ -1,0 +1,17 @@
+require "pathname"
+require "json"
+
+module DiasporaFederation
+  # A helper class to access the JSON schema.
+  module Schemas
+    # federation_entities schema uri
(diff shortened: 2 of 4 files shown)


diaspora_federation.gemspec
```diff
@@ -17,7 +17,8 @@ Gem::Specification.new do |s|
   s.license     = "AGPL-3.0"
 
   s.files       = Dir["lib/**/*", "LICENSE", "README.md", "Changelog.md"] -
-                    Dir["lib/diaspora_federation/{engine,rails,test}.rb",
+                    Dir["lib/diaspora_federation/{engine,rails,schemas,test}.rb",
+                        "lib/diaspora_federation/schemas/*",
```
cmrd-senya: Why do you add schemas to the `diaspora_federation` gem while we have them in `diaspora_federation-json_schema`?

SuperTux88: This is the exclude list. I noticed that it was accidentally added to the main gem.

cmrd-senya: Ah, I see. This `-` is hardly noticable

lib/diaspora_federation/schemas.rb
```diff
@@ -0,0 +1,17 @@
...
+module DiasporaFederation
+  # A helper class to access the JSON schema.
+  module Schemas
+    # federation_entities schema uri
+    FEDERATION_ENTITIES_URI = "https://diaspora.github.io/diaspora_federation/schemas/federation_entities.json".freeze
+
+    # Parsed federation_entities schema
+    def self.federation_entities
+      @federation_entities ||= JSON.parse(
+        Pathname.new(__dir__).join("schemas", "federation_entities.json").expand_path.read
```
cmrd-senya: What difference does `expand_path` makes here?

SuperTux88: It returns the absolute path, but I only copied it from [here](https://github.com/diaspora/diaspora/blob/7934c1e969729686bcc964948bac3a381273ae03/lib/node_info.rb#L157-L159). But you're right, it doesn't make a difference here with `.read`, so I removed it.

cmrd-senya:
Merged, thank you! :cookie:

PR #73, opened 2017-08-25
Allow reshares with no root

cmrd-senya:
Follow-up for https://discourse.diasporafoundation.org/t/0-7-account-export-error/874

### DIFF
--- a/lib/diaspora_federation/entities/reshare.rb
+++ b/lib/diaspora_federation/entities/reshare.rb
@@ -31,6 +31,11 @@ module DiasporaFederation
       # Fetch and receive root post from remote, if not available locally
       # and validates if it's from the correct author
       def validate_root
+        return if root_author.nil? && root_guid.nil?
+
+        raise Entity::ValidationError, "#{self}: root_guid can't be nil if root_author is present" if root_guid.nil?
+        raise Entity::ValidationError, "#{self}: root_author can't be nil if root_guid is present" if root_author.nil?
+
         root = RelatedEntity.fetch(root_author, "Post", root_guid)
 
         return if root_author == root.author
--- a/spec/lib/diaspora_federation/entities/reshare_spec.rb
+++ b/spec/lib/diaspora_federation/entities/reshare_spec.rb
@@ -74,6 +74,31 @@ XML
             Entities::Reshare.from_xml(Nokogiri::XML(xml).root)
           }.to raise_error Entity::ValidationError
         end
+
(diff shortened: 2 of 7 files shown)


spec/support/shared_validator_specs.rb
```diff
@@ -124,6 +128,17 @@ def alphanumeric_string(length)
   end
 end
 
+shared_examples "a nilable guid validator" do
```
SuperTux88: Can you please (in a second commit) use this for [photo_validator_spec.rb](https://github.com/diaspora/diaspora_federation/blob/e26a86c173a7fdd222bd3cba677f03cf08c6487c/spec/lib/diaspora_federation/validators/photo_validator_spec.rb#L19) too?

spec/lib/diaspora_federation/entities/reshare_spec.rb
```diff
@@ -74,6 +74,31 @@ module DiasporaFederation
...
           }.to raise_error Entity::ValidationError
         end
+
+        it "validates a reshare with no root" do
+          data[:root_author] = nil
+          data[:root_guid] = nil
+
+          expect {
+            Entities::Reshare.from_xml(Nokogiri::XML(xml).root)
+          }.not_to raise_error
```
SuperTux88: Instead of `not_to raise_error` I would rather check if the parsed `root_author` and `root_guid` is `nil`. The "not to raise error"-part is implicit then.

lib/diaspora_federation/entities/reshare.rb
```diff
@@ -31,6 +31,11 @@ def to_s
       # Fetch and receive root post from remote, if not available locally
       # and validates if it's from the correct author
       def validate_root
+        return if root_author.nil? && root_guid.nil?
+
+        raise Entity::ValidationError, "root_guid can't be nil if root_author is present" if root_guid.nil?
+        raise Entity::ValidationError, "root_author can't be nil if root_guid is present" if root_author.nil?
```
SuperTux88: I would like to have the `#{self}` somewhere in the error message. It's a lot easier to grep the guid when somebody asks "why didn't you receive this message. I already had a some questions from other developers from other software implementing the protocol, why some messages failed. (Even when this error shouldn't happen often)

So maybe just add it to the beginning: `"#{self}: ..."`

cmrd-senya:
Updated

SuperTux88:
Merged, thank you 🍪

PR #74, opened 2017-08-27
Parse RFC 7033 webfinger

SuperTux88:
Second part for diaspora/diaspora#5963

### DIFF
--- a/docs/discovery/webfinger.md
+++ b/docs/discovery/webfinger.md
@@ -14,8 +14,6 @@ GET /.well-known/webfinger
 
 Let's assume we are searching for `alice@example.org`, then we need to make a request to `example.org`.
 
-The request should first be tried with https, if this doesn't work, the requesting server should fallback to http.
-
 #### Parameters
 
 | Name       | Description                                                  |
--- a/docs/discovery/webfinger.md
+++ b/docs/discovery/webfinger.md
@@ -2,88 +2,43 @@
 title: WebFinger
 ---
 
-diaspora\* uses an old draft of [WebFinger][webfinger-draft] to discover users from other pods.
-
-{% include warning_box.html
-   title="Old WebFinger"
-   content="<p>diaspora* doesn't yet fully support the RFC 7033 WebFinger!</p>"
-%}
-
-## WebFinger endpoint discovery
-
-To find the WebFinger endpoint, the requesting server must get the [host metadata][host-meta] information for the
-domain of the searched diaspora\* ID.
-
-### Request
-
-Let's assume we are searching for `alice@example.org`, then we need the host metadata for `example.org`.
-
-The request should first be tried with https, if this doesn't work, the requesting server should fallback to http.
-
-If the server response indicates that the host-meta resource is located elsewhere (a 301, 302, or 307 response status
-code), the requesting server should try to obtain the resource from the location provided in the response.
-
-#### Example
-
-~~~
-GET /.well-known/host-meta
-Host: example.org
-~~~
-
-### Response
-
-The host-meta document must be in the [XRD 1.0 document format][xrd] and should be served with the
-`application/xrd+xml` media type.
-
-#### Mandatory Link Relations
-
-The Host Metadata response must contain the following link relations:
-
-| Link Relation | Type         | Description                                                               |
-| ------------- | ------------ | ------------------------------------------------------------------------- |
-| lrdd          | url template | The template to the URL where the server provides the WebFinger endpoint. |
-
-The Host Metadata response may contain other optional link relations.
-
-#### Example
-
-~~~
-Status: 200 OK
-Content-Type: application/xrd+xml; charset=utf-8
-~~~
-~~~xml
-<?xml version="1.0" encoding="UTF-8"?>
-<XRD xmlns="http://docs.oasis-open.org/ns/xri/xrd-1.0">
-  <Link rel="lrdd" type="application/xrd+xml" template="https://example.org/.well-known/webfinger.xml?resource={uri}"/>
-</XRD>
-~~~
+diaspora\* uses [WebFinger][webfinger-rfc] to discover users from other pods.
 
 ## WebFinger
 
 ### Request
 
+~~~
+GET /.well-known/webfinger
+~~~
+
+Let's assume we are searching for `alice@example.org`, then we need to make a request to `example.org`.
+
+The request should first be tried with https, if this doesn't work, the requesting server should fallback to http.
+
 #### Parameters
 
-The requesting server must replace the ``{uri}`` in the lrdd-template from the host-meta request with the diaspora\* ID
-of the searched person.
+| Name       | Description                                                  |
+| ---------- | ------------------------------------------------------------ |
+| `resource` | The "acct" URI for the diaspora\* ID of the searched person. |
 
 #### Example
 
 ~~~
-GET /.well-known/webfinger.xml?resource=acct:alice@example.org
+GET /.well-known/webfinger?resource=acct:alice@example.org
 Host: example.org
 ~~~
 
 ### Response
 
-The webfinger document must be in the [XRD 1.0 document format][xrd] and should be served with the
-`application/xrd+xml` media type.
+The WebFinger document must be in the [JSON Resource Descriptor (JRD) format][jrd] and should be served with the
+`application/jrd+json` media type.
 
 If the requested diaspora\* ID is unknown by the server, it must return a 404 status code.
 
 #### Subject
 
-The ``Subject`` element should contain the webfinger address that was asked for. If it does not, then this webfinger
+The `subject` element should contain the WebFinger address that was asked for. If it does not, then this WebFinger
 profile must be ignored.
 
 #### Mandatory Link Relations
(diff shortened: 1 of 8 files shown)


spec/lib/diaspora_federation/discovery/discovery_spec.rb
```diff
@@ -88,66 +91,148 @@ module DiasporaFederation
...
 
         expect_callback(:save_person_after_webfinger, kind_of(Entities::Person))
-        person = Discovery::Discovery.new(account).fetch_and_save
+        person = subject.fetch_and_save
 
         expect(person.guid).to eq(alice.guid)
         expect(person.diaspora_id).to eq(account)
       end
 
       it "falls back to http if https fails with ssl error" do
```
cmrd-senya: Similar to the above, falling back to https we actually allow MitM attack. If :imp: wants to serve something instead of the original resource he just replies to the https request with self-signed certificate and then just passes desired data via http.

spec/lib/diaspora_federation/discovery/discovery_spec.rb
```diff
@@ -88,66 +91,148 @@ module DiasporaFederation
           callback_person = person
         end
 
-        expect(Discovery::Discovery.new(account).fetch_and_save).to be(callback_person)
+        expect(subject.fetch_and_save).to be(callback_person)
       end
 
       it "falls back to http if https fails with 404" do
```
cmrd-senya: Does it make sense to fall back to HTTP in case if the same URI at HTTPS replied as `404`? We expect to have the same resource for HTTP and HTTPS at the same address, so if HTTPS is `404` and HTTP is not then it's likely to be some kind of evil :imp:.

Also, RFC 7033 says
> A WebFinger request is an HTTPS request to a WebFinger resource.  A
>   WebFinger resource is a well-known URI [3] using the HTTPS scheme
>   constructed along with the required query target and optional link
>   relation types.  WebFinger resources MUST NOT be served with any
>   other URI scheme (such as HTTP).

which we clearly violate this way. Is there any special reason why we need the HTTP fallback at all?

SuperTux88: We always did it that way. We allow pods to run on http (while we strongly recommend https!), so we need to allow to fetch webfinger via http. The 404 case maybe doesn't make much sense, but there are webservers configured to only work with http and 404 everything on https (I saw a pod like this once). We maybe need to discuss if we want to block such misconfigurations (and enforce https everywhere), but currently we allow that (and did always in the past).

spec/lib/diaspora_federation/discovery/discovery_spec.rb
SuperTux88: Same as above, it was always that way. But it's only for discovery, once a contact is known, there is no possibility to change the key or force it back to http when it was once https.

If you think we should change something here it's probably the best when you open a discussion on discourse. But I think it's out of scope here (because as said, it's currently that way, and we would at least need a major release to remove http compatibility)

SuperTux88:
I fixed the other two things and updated the PR.

spec/lib/diaspora_federation/discovery/discovery_spec.rb
zotlabs: What we did in this case was fallback on XML/XRD for non-SSL sites. That way everything still works and it's spec-compliant, although it does mean maintaining the XRD service.  I notice a couple of projects that generate both the json webfinger and the personal xrd from a common backend structure which would relieve a lot of the maintenance problem - if you see that as a problem.

cmrd-senya: > We always did it that way. 

But we didn't have RFC 7033 support before and RFC 7033 support requires to use HTTPS. So in order to be spec compiant we shouldn't provide this WebFinger route when pod doesn't have HTTPS available, and therefore shouldn't fetch it via HTTP.

You introduce RFC 7033 in this PR that's why it is in scope of this PR.

So I think the right thing to do is to support only the old draft for HTTP pods. It explicitly allows this:
>   WebFinger itself is agnostic regarding the scheme of such a URI: it
>   could be an "acct" URI as defined in the next section, an "http" or
>   "https" URI, a "mailto" URI, or some other scheme.

If we ever want to drop the old WebFinger, then we'll have to consider dropping HTTP support at all.

Also the draft talks about signatures for HTTP:
>   When using HTTP to request an XRD document, WebFinger clients SHOULD
>   verify the XRD document's signature, if present, to ensure that the
>   XRD document has not been modified.  Additionally, WebFinger servers
>   SHOULD include a signature for XRD documents served over HTTP.

But I can't find the signing procedure described, mainly, how to perform key exchange, which doesn't make much sense to do over the same HTTP channel (that's why it is probably not in the final spec).

What do you think?

Flaburgan: If we drop the "old WebFinger", we have to do it in a major release. So it makes sense to me to drop HTTP support in the same time. (Something we should probably do, now that browsers dislike http a lot and let's encrypt make damn easy to set up https).

So I would go for "https only for the "new" webfinger, if https is not available then fallback to the old one".

SuperTux88: If we want to remove http support we can add the RFC 7033 WebFinger without http support, and fallback to the old WebFinger if needed. But then we should drop http support at the next major (so drop old WebFinger support with it, at least stop fetching it), otherwise we win absolutely no security if it is still possible to downgrade to the old way without https. But then we should start warning podmins now, so they have time to add https if they don't have it already.

What should we do for development? I think we should at least allow http fallback for development, because we don't have https there. Keeping the old WebFinger for development but using the RFC 7033 WebFinger in production doesn't sound like a good idea to me.

cmrd-senya: Maybe we should make the http fallback for RFC 7033 configurable and off by default? So by default we don't fall back to HTTP for the new WebFinger and if the `diaspora_federation` gem based application opts in, then we do fall back. We can even set it to fallback in diaspora itself (override the default), but for `diaspora_federation` it will be off by default, so our library is RFC 7033 compliant by default with no exceptions. For diaspora we could allow this fallback either for both production and development, or for development only, or we can have it on until we support http and turn it off at the time we drop http whenever this happens and leave it just for development purposes.

cmrd-senya: That is how it works for OpenID Connect gems I worked with: you can use it with HTTP, until you explicitly set some magic setting that tells the gem to ignore https only restriction.

SuperTux88: OK, so I'll add a setting for the fallback, and then we can handle that in diaspora.

SuperTux88: Done and PR updated.

docs/discovery/webfinger.md
```diff
@@ -2,88 +2,43 @@
...
-  <Link rel="lrdd" type="application/xrd+xml" template="https://example.org/.well-known/webfinger.xml?resource={uri}"/>
-</XRD>
+GET /.well-known/webfinger
 ~~~
 
-## WebFinger
+Let's assume we are searching for `alice@example.org`, then we need to make a request to `example.org`.
 
-### Request
+The request should first be tried with https, if this doesn't work, the requesting server should fallback to http.
```
cmrd-senya: This should be updated according to our decion upon http. We shouldn't say "the requesting server should fallback to http". Instead we can say that by default we support https only (since we follow the RFC 7033), but a protocol implementation can optionally support `http` schema fallback which is not RFC 7033 compliant, but can be useful in some cases.

lib/diaspora_federation.rb
```diff
@@ -74,6 +75,18 @@ class << self
     #   @param [String] value path to certificate authorities
     attr_accessor :certificate_authorities
 
+    # Configure if RFC 7033 WebFinger discovery should fallback to http if https fails (default: +false+)
```
cmrd-senya: Nit: I would change this to "Configure if  ~RFC 7033~ WebFinger discovery should fallback to http if https fails", since strictly saying falling back to http is not compliant with RFC 7033. To make it clear that we talk only about the "new" WebFinger we can add another comment line saying:

```ruby
# This setting only applies to the WebFinger route from RFC 7033 `/.well-known/webfinger`. Legacy WebFinger flow unconditionally falls back to http.
```

spec/lib/diaspora_federation/discovery/discovery_spec.rb
```diff
@@ -146,6 +116,84 @@ module DiasporaFederation
...
+
+            expect(person.guid).to eq(alice.guid)
+            expect(person.diaspora_id).to eq(account)
+          end
+
+          it "falls back to legacy WebFinger if https fails with ssl error" do
+            stub_request(:get, "https://localhost:3000/.well-known/webfinger?resource=acct:#{account}")
+              .to_raise(OpenSSL::SSL::SSLError)
+            stub_request(:get, "https://localhost:3000/.well-known/host-meta")
+              .to_return(status: 200, body: host_meta_xrd)
```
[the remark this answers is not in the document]
SuperTux88: I did that now, but now this test is exactly the same as the test below for legacy WebFinger. But since it still tests what it should test it doesn't matter, and it will change once legacy WebFinger is removed.

spec/lib/diaspora_federation/discovery/discovery_spec.rb
```diff
@@ -88,66 +91,192 @@ module DiasporaFederation
...
       end
 
-      it "fails if the diaspora* ID was not found" do
-        stub_request(:get, "https://localhost:3000/.well-known/host-meta")
-          .to_return(status: 200, body: host_meta_xrd)
-        stub_request(:get, "http://localhost:3000/.well-known/webfinger.xml?resource=acct:#{account}")
-          .to_return(status: 404)
+      context "legacy WebFinger" do
+        it "falls back to legacy WebFinger" do
+          stub_request(:get, "https://localhost:3000/.well-known/webfinger?resource=acct:#{account}")
```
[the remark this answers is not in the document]
SuperTux88: No, I wanted to test the different scenarios here.

And older pods (since 0.6.0.0) did have RFC 7033 WebFinger (for OpenID), just without any usable data, so the "real world scenario" would be a `ValidationError` because of the missing data, not a 404. So I changed the `"foobar"` (which raised a `JSON::ParserError`) to the JSON an old pod will return, to simulate the most common scenario.

But I still want to test `404` and `SSLError` here.

docs/discovery/webfinger.md
SuperTux88: Good catch. I just removed the sentence completely. If it is in the RFC, we don't need to repeat it here. I only described the most basic WebFinger here and didn't copy/repeat the whole RFC. It's only to mention what parts in the response are required for us, everything else is in the RFC.

And I also don't think we need to allow people to break the RFC here. If people do that (for whatever reason, for us it will be development mode), then they can do it.

cmrd-senya:
Merged, thanks!

PR #76, opened 2017-08-28
Make height and width optional for photos

SuperTux88:
This is needed to fix profile exports, see [here](https://discourse.diasporafoundation.org/t/0-7-account-export-error/874/2?u=supertux88).


SuperTux88:
You are right, it's a good idea to use the new `optional` flag. There is a special case for arrays where it needs to be an empty array (an not nil) even when it's optional (`photos` in `StatusMessage`), so I added a check to not ignore `not_nil` rules.

I also added a commit to removed the `nilable` parameter from the `guid` rule, because we can use the `optional` flag there too.

cmrd-senya:
Merged, thanks! :cookie: :cookie: :cookie:

PR #77, opened 2017-08-31
Improve error handling for discovery errors

SuperTux88:
Always raise a `DiscoveryError` when something with the discovery fails


SuperTux88:
Merged, thanks for the review @denschub 🍪

Issue #75, opened 2017-08-27
diaspora:// scheme for pod independent links


SuperTux88:
On diaspora we have the problem that people want to link to another post, but when I just link to a post like this `https://nerdpol.ch/posts/3919682` most people can't comment on that post, because it links to the wrong post. Also: When the post is private, a person from another pod can't see the post, even when the person does have permissions to view that post on their pod.

People try to workaround this with creating relative URLs with the GUID like this `/posts/17faf230675101350d995254001bd39e` (there are even userscripts to automate this). This is better in some ways: When it's a public post or a private post where I have permissions, I can see it on my pod and comment or like. But it has still problems:
* When my pod doesn't know the post yet, I only see a 404 (that was better with the direct link to another pod where the post was already known)
* This only works with diaspora, because every other software compatible with the diaspora protocol has other routes to display posts.

I suggest to introduce a diaspora:// URI scheme to link to another post. The links should have the same format as for [fetching entities](https://diaspora.github.io/diaspora_federation/federation/fetching.html):
```
diaspora://:author/:type/:guid
```
Example:
```
diaspora://hq@pod.diaspora.software/post/17faf230675101350d995254001bd39e
```

That way it would be possible to fetch linked entities (when it is public), that solves the first problem above. And every software compatible to the protocol can convert these links to the internally used route to display the entity (without guessing internal links of other software), that solves the second problem above.

Diaspora currently only supports to fetch posts, so it would also only be possible to link to other posts in the beginning. But that is extendable to other entity-types in the future.

I wanted to add forward compatibility in diaspora 0.7.1.0 (ability to handle and display diaspora:// URLs correctly in the frontend), so we can start using them in diaspora 0.8.0.0 and replace pod-internal URLs with pod independent URLs.

SuperTux88:
I added the `author` to the diaspora:// URL above. This has two advantages:
1. We can validate, that if we already know a post with that GUID, it is the post from the author the person linking to it intended to link (the same as we validate `root_author` for a reshare).
2. We can fetch it from the author directly instead of needing to redirect from the senders pod.

Flaburgan:
Is author mandatory or optional?

SuperTux88:
Since it's at the beginning if the URL, it's mandatory. But every object has an author, and I think it's important to validate that.

SuperTux88:
The protocol is now registered in the IANA URI Schemes registry \o/ (Thanks @denschub for the work 🍪)

[...]

PR #78, opened 2017-09-04
Fetch posts linked with diaspora:// URLs

SuperTux88:
This adds support to fetch linked entities from markdown text if they aren't already known to the server.

This is the first step needed for #75.


SuperTux88:
I rebased and resolved the conflicts :)

PR #79, opened 2017-09-04
Allow to receive non-public profiles without private data

SuperTux88:
This is needed to send non-public profiles to a public endpoint when not sharing with a person on the target pod. This is needed to fix diaspora/diaspora#2884.

I already described that more in [this comment](https://github.com/diaspora/diaspora/pull/7501#issuecomment-317245346) where this PR is needed for these two scenarios:

> * private details + searchable:
>     * send to all contacts the person shares with privately
>     * **send public profile (without private parts) to all other known pods publicly**
> * private details + not searchable:
>     * send to all contacts the person shares with privately
>     * **send public profile to all other contacts publicly**

### DIFF
--- a/docs/_entities/profile.md
+++ b/docs/_entities/profile.md
@@ -4,6 +4,21 @@ title: Profile
 
 This entity contains all the profile data of a person.
 
+The profile consists of two parts. The first is the base profile with the name, 
+the avatar and the tags of the person. This part is always public and visible to
+everyone. The boolean flags (`searchable`, `public` and `nsfw`) are metadata and
+public too.
+
+The second part is the extended profile consisting of `bio`, `birthday`, `gender`
+and `location` is not public by default (and only visible to contacts the person
+shares with). The owner of the profile can decide if this information should be
+public (that's what the `public` flag is for) and then the extended profile is
+visible to non-contacts too.
+
+Because of that there is a special case where the `public`-flag is `false`, but
+the profile was received via the public route. In this case the profile should
+only contain the base profile.
+
 ## Properties
 
 | Property | Type                         | Description                      |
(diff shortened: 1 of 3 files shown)


SuperTux88:
I updated the documentation to explain the `public` flag and the base and extended profile.

cmrd-senya:
> private details + searchable:
> - send public profile (without private parts) to all other known pods publicly
>
> private details + not searchable:
> - send public profile to all other contacts publicly

Given a user changes their profile from searchable to non-searchable. Given there is a remote pod which had fetched this user's profile before but at this pod this user doesn't have any contacts. In this case does it mean that the remote pod will "forever" display the last version of user's public profile that was pushed there? Or at least until the user changes it back to searchable? Does it mean, that users in this case won't have any control on public data which such remote pods will display for their profile (which can be accessed by their guid)? So they can't delete their information that they don't want to be displayed there anymore?

If I got it right, then I guess we need some way to inform the remote pod about the fact that we will not send updates anymore. Possible way could be a `Retraction` message with `Profile` type. For the recepient pod the meaning of this message is "Stop showing this profile's public and private details because they are outdated". So basically we need to clear people's profile if they go "non-searchable" at the pods where they don't have contacts. And since there is no way to know if the remote pod will share public profile updates to us we need it to report the `Retraction` explicitly.

What do you think, @SuperTux88?

SuperTux88:
>  In this case does it mean that the remote pod will "forever" display the last version of user's public profile that was pushed there?

Yes, for non-searchable profiles it is still the same as it already is today, but with this change we will improve it for searchable profiles, while keeping it the same as now for non-searchable profiles.

> which can be accessed by their guid

The profile can only be accessed for users on that pod, it's not public there. However the name and avatar is displayed at posts which are publicly accessible.

> If I got it right, then I guess we need some way to inform the remote pod about the fact that we will not send updates anymore.

And then? What should the pod do? Delete the profile? That's not possible if there are already posts/comments with this profile. Or remove name and avatar and use diaspora ID and default avatar instead? Then we should change WebFinger to never include a name or an avatar, because when a pod fetches a profile with WebFinger the profile will never be updated too and the pod will never receive a `Retraction`.

There is simply no solution for that, when a pod fetched the profile but doesn't receive any updates for a profile he doesn't receive any updates ... it's the same as now ... The only reason for this is to improve it for searchable profiles by pushing more updates than now.

Because profiles are currently searchable by default I wanted to add a setting for that to the getting started page, so new users can decide that after registration, before the profile is fetched/federated. But that is out of scope here and can be discussed when this is added to diaspora.

SuperTux88:
The pod could fetch the profile every time (which would make everything slow), or cache it only for limited time and then WebFinger it again for updates ... but that's everything out of scope here ...

SuperTux88:
PR updated.

cmrd-senya:
Merged, thanks!

PR #81, opened 2017-09-05
Relay likes for comments

SuperTux88:
This fixes relaying of likes of comments, needed for diaspora/diaspora#2999

### DIFF
--- a/docs/federation/relayable.md
+++ b/docs/federation/relayable.md
@@ -2,10 +2,14 @@
 title: Relayable
 ---
 
-If a person participates on an entity, it needs to be relayed via the author of the parent entity, because only the
-parent author knows, to whom they shared the original entity.
+If a person participates on an entity, it needs to be relayed via the author of the root entity, because only the
+root author knows, to whom they shared the original entity.
 
-Such entities are:
+The root is the most top parent of the relayable: If the parent of a relayable is a relayable itself,
+the parent of that should be used, until it isn't a relayable anymore.
+For example, for both, a `Like` for a `Post` and a `Like` for a `Comment`, the author of the `Post` should be used.
+
+Such relayable entities are:
 
 * [Comment][comment]
 * [Like][like]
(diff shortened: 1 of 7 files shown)


lib/diaspora_federation/entities/related_entity.rb
```diff
@@ -24,6 +24,14 @@ class RelatedEntity < Entity
       #   @return [RelatedEntity] parent entity
       entity :parent, Entities::RelatedEntity, default: nil
 
+      # The most top entity, this entity is responsible for relaying relayables
+      # @return [RelatedEntity] absolute parent entity
+      def top_entity
```
Flaburgan: Not sure "Top entity" is the best wording here, maybe something like "relayable responsible"? I don't know, but something containing the meaning of what we're doing that here.

SuperTux88: I'm not sure too. `top_entity` is short and says what it is. The documentation above the method definition describes for what it's used with more words. Before `parent` was used instead of `relayable_responsible`, even when it was what it was doing before.

denschub: `root_entity` maybe?

SuperTux88: I also thought about `root_...` but I thought it could be confusing with `root_...` at reshares :/

denschub: Yeah, I thought so as well. However, `root` is actually a pretty accurate word, since `root` is always the top level. Eventually, this will also be true for reshares, so maybe, reusing the same terminology is a good way to go?

SuperTux88: I renamed it to `root`. While it's maybe a bit weird now, because we have two things named `root`, I think since reshares won't have comments and likes (relayables) in the future, this will not create conflicts there in the future.

spec/lib/diaspora_federation/entities/retraction_spec.rb
```diff
@@ -83,6 +83,36 @@ module DiasporaFederation
...
+            :related_entity,
+            author: Fabricate.sequence(:diaspora_id),
+            parent: Fabricate(:related_entity, author: alice.diaspora_id)
+          )
+        }
+        let(:relayable_target) { Fabricate(:related_entity, author: bob.diaspora_id, parent: comment) }
+        let(:relayable_data) { data.merge(target_type: "Like", target: relayable_target) }
+
+        it "allows target author" do
+          entity = Entities::Retraction.new(relayable_data)
```
cmrd-senya: Nit: replace `entity = Entities::Retraction.new(relayable_data)` with a `let` block?

docs/federation/relayable.md
```diff
@@ -11,6 +11,9 @@ Such entities are:
 * [Like][like]
 * [PollParticipation][poll_participation]
 
+If the parent is a relayable itself, the parent of that should be used, until it isn't a relayable anymore.
```
cmrd-senya: I feel that explanations in this section may be confusing.

Maybe we should change the text at lines 5-6 to use the word `root` instead of the word `parent`? And then we can explain what the `root` word means: it is a top parent for entity, which is normally a `Post` for both `Like` for a `Post` and `Like` for a `Comment`.

Also I guess we should replace `parent` with `root` everywhere in this document where we actually mean `root`. 

E.g. line 33: `The parent author then must envelop it in a new [Magic Envelope][magicsig] and send the entity to all the recipients of the parent entity.`. Here we actually mean `root`, becase for `Comment` for `Like` it is not the parent author, but the root author who should perform the relaying.

SuperTux88: Sure, you're right, I fixed that.

cmrd-senya:
Merged, thanks!

PR #82, opened 2017-09-06
Add text property to reshares to allow quoted reshares

SuperTux88:
With this property the author of the reshare can add his own thoughts about the post to the reshare.

@jaywink mentioned that some time ago that he wants to use this property and I thought it is a good idea to add it to add it to the official documentation.

This will also help to solve diaspora/diaspora#4219 by adding quoted reshares to diaspora as discussed [here](https://discourse.diasporafoundation.org/t/reshares/357).

### DIFF
--- a/docs/_entities/reshare.md
+++ b/docs/_entities/reshare.md
@@ -23,10 +23,11 @@ is not valid.
 
 ## Optional Properties
 
-| Property                | Type (Length)          | Description                                                                                |
-| ----------------------- | ---------------------- | ------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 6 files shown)


SuperTux88:
As discussed via IRC, the `reshare` message should stay reshare-only and shouldn't be used for quoting.

It needs to be discussed if we need a new `quote` message (copy of the current reshare message with a text property), or if we just use the `status_message` message and link the quoted post there with a diaspora:// URL (#75) and start embedding posts that are linked.

cmrd-senya:
The discussion on that topic is expected to take place at https://discourse.diasporafoundation.org/t/allowing-people-to-easier-link-to-reference-posts-in-their-own-posts/898

PR #87, opened 2017-09-22
Don't fetch the same entity twice in the same thread

SuperTux88:
Normally this shouldn't happen, but it could be used as an attack vector when somebody creates this scenario manually and links to it, so pods start fetching the loop.

When two threads fetch the same entity (which can happen when receiving the same entity twice), this isn't a problem, because the receive method can handle duplicates anyway.

### DIFF
--- a/lib/diaspora_federation/federation/fetcher.rb
+++ b/lib/diaspora_federation/federation/fetcher.rb
@@ -7,15 +7,9 @@ module DiasporaFederation
       # @param [Symbol, String] entity_type snake_case version of the entity class
       # @param [String] guid guid of the entity to fetch
       def self.fetch_public(author, entity_type, guid)
-        url = DiasporaFederation.callbacks.trigger(
-          :fetch_person_url_to, author, "/fetch/#{entity_name(entity_type)}/#{guid}"
-        )
-        response = HttpClient.get(url)
-        raise "Failed to fetch #{url}: #{response.status}" unless response.success?
-
-        magic_env_xml = Nokogiri::XML(response.body).root
-        magic_env = Salmon::MagicEnvelope.unenvelop(magic_env_xml)
-        Receiver::Public.new(magic_env).receive
+        type = entity_name(entity_type).to_s
+        raise "Already fetching ..." if fetching[type].include?(guid)
+        fetch_from_url(author, type, guid)
       rescue => e # rubocop:disable Lint/RescueWithoutErrorClass
         raise NotFetchable, "Failed to fetch #{entity_type}:#{guid} from #{author}: #{e.class}: #{e.message}"
       end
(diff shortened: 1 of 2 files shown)


lib/diaspora_federation/federation/fetcher.rb
```diff
@@ -7,15 +7,9 @@ module Fetcher
...
-        raise "Failed to fetch #{url}: #{response.status}" unless response.success?
-
-        magic_env_xml = Nokogiri::XML(response.body).root
-        magic_env = Salmon::MagicEnvelope.unenvelop(magic_env_xml)
-        Receiver::Public.new(magic_env).receive
+        type = entity_name(entity_type).to_s
+        raise "Already fetching ..." if fetching[type].include?(guid)
+        fetch_from_url(author, type, guid)
       rescue => e # rubocop:disable Lint/RescueWithoutErrorClass
         raise NotFetchable, "Failed to fetch #{entity_type}:#{guid} from #{author}: #{e.class}: #{e.message}"
```
cmrd-senya: Nit: for me it would feel more accurate, if we created a `AlreadyFetching` exception class and propagated it up instead of raising `NotFetchable` for this specific case. What do you think, @SuperTux88 ?

SuperTux88: No, I do this intentionally, because the only thing `fetch_public` can raise is a `NotFetchable` error. So the only place we need to catch everything when something fails with fetching is here, and everything can just handle `NotFetchable`. The details are logged.

I do the same thing with `DiscoveryError` for when something with the discovery fails.

Yes, I could extend `AlreadyFetching` from `NotFetchable`, so it could be catched with `NotFetchable`. but I don't see what the advantage is when you have a separate class for this case?

cmrd-senya: Well, I mostly disliked the working, since it is not `NotFetchable` when we're already fetching it. But since this exception is public API, then I guess we shouldn't change it for such minor reasons.

lib/diaspora_federation/federation/fetcher.rb
```diff
@@ -28,6 +22,24 @@ def self.fetch_public(author, entity_type, guid)
         class_name.gsub(/(.)([A-Z])/, '\1_\2').downcase
       end
 
+      private_class_method def self.fetch_from_url(author, type, guid)
+        fetching[type] << guid
+        url = DiasporaFederation.callbacks.trigger(:fetch_person_url_to, author, "/fetch/#{type}/#{guid}")
+        response = HttpClient.get(url)
+        raise "Failed to fetch #{url}: #{response.status}" unless response.success?
+
+        magic_env_xml = Nokogiri::XML(response.body).root
```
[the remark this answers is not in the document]
SuperTux88: You are right, it's exactly the same (with the exception that it has duplicate logging now, in case of a failure), so I changed it.

cmrd-senya:
Merged, thanks!

lib/diaspora_federation/federation/fetcher.rb
cmrd-senya: BTW perhaps this method description needs `@raise NotFetchable` in the description. But I already merged the PR, so whatever :)

SuperTux88: I added that now directly to develop :)

PR #89, opened 2017-10-30
Extract singning of AccountMigration to a different module and introduce alternative form for account migration message

cmrd-senya:
As discussed at https://github.com/diaspora/diaspora/issues/908, if we want to make `AccountMigration` object fetchable from the new pod and at the same time we don't want to keep private key of the old person on the new pod, we need to save signature of the old person.

In order to do that, we need two things:
- Allow `AccountMigration` model from diaspora* to compute a signature without instantiating the entity, so we can compute the signature and store it to the database at the time when we build the `AccountMigration` object at the new home pod (with old private key as user input).
- Allow alternative form of `AccountMigration` entity where `author` is the new user and old user is defined by a separate parameter.

This PR introduces these changes.

In [this branch](https://github.com/cmrd-senya/diaspora/commits/resend-account-close-event) I develop changes in diaspora which make use of these changes, but only of the first part: it saves signature for `AccountMigration` to the DB. Also in that branch I resend `AccountDeletion`/`AccountMigration` to the sender if we received a message for a closed account. I'll prepare that branch as a PR soon.

### DIFF
--- a/lib/diaspora_federation/entities/account_migration.rb
+++ b/lib/diaspora_federation/entities/account_migration.rb
@@ -5,7 +5,7 @@ module DiasporaFederation
     #
     # @see Validators::AccountMigrationValidator
     class AccountMigration < Entity
-      include Signable
+      include AccountMigration::Signable
 
       # @!attribute [r] author
       #   The old diaspora* ID of the person who changes their ID
@@ -23,16 +23,18 @@ module DiasporaFederation
       #   @return [String] signature
       property :signature, :string, default: nil
 
-      # @return [String] string representation of this object
-      def to_s
-        "AccountMigration:#{author}:#{profile.author}"
+      def old_user_id
+        author
       end
+
+      # @return [String] string representation of this object
+      alias to_s unique_migration_descriptor
 
       # Shortcut for calling super method with sensible arguments
       #
       # @see DiasporaFederation::Entities::Signable#verify_signature
       def verify_signature
-        super(profile.author, :signature)
+        super(signer_id, :signature)
       end
 
       # Calls super and additionally does signature verification for the instantiated entity.
@@ -44,9 +46,12 @@ module DiasporaFederation
 
       private
 
-      # @see DiasporaFederation::Entities::Signable#signature_data
-      def signature_data
-        to_s
+      def new_user_id
+        profile.author
+      end
+
+      def signer_id
+        new_user_id
       end
 
       def enriched_properties
@@ -59,10 +64,10 @@ module DiasporaFederation
       # @raise [NewPrivateKeyNotFound] if the new user's private key is not found
       # @return [String] A Base64 encoded signature of #signature_data with key
       def sign_with_new_key
-        privkey = DiasporaFederation.callbacks.trigger(:fetch_private_key, profile.author)
-        raise NewPrivateKeyNotFound, "author=#{profile.author} obj=#{self}" if privkey.nil?
+        privkey = DiasporaFederation.callbacks.trigger(:fetch_private_key, signer_id)
+        raise NewPrivateKeyNotFound, "signer=#{signer_id} obj=#{self}" if privkey.nil?
         sign_with_key(privkey).tap do
-          logger.info "event=sign status=complete signature=signature author=#{profile.author} obj=#{self}"
+          logger.info "event=sign status=complete signature=signature signer=#{signer_id} obj=#{self}"
         end
       end
 
--- a/lib/diaspora_federation/entities/account_migration.rb
+++ b/lib/diaspora_federation/entities/account_migration.rb
@@ -8,7 +8,9 @@ module DiasporaFederation
       include AccountMigration::Signable
 
       # @!attribute [r] author
-      #   The old diaspora* ID of the person who changes their ID
+      #   Sender of the AccountMigration message. Usually it is the old diaspora* ID of the person who changes their ID.
+      #   This property is also allowed to be the new diaspora* ID, which is equal to the author of the included
+      #   profile.
       #   @see Person#author
       #   @return [String] author diaspora* ID
       property :author, :string
@@ -19,12 +21,29 @@ module DiasporaFederation
       entity :profile, Entities::Profile
 
       # @!attribute [r] signature
-      #   Signature that validates original and target diaspora* IDs with the new key of person
+      #   Signature that validates original and target diaspora* IDs with the private key of the second identity, other
+      #   than the entity author. So if the author is the old identity then this signature is made with the new identity
+      #   key, and vice versa.
       #   @return [String] signature
       property :signature, :string, default: nil
 
-      def old_user_id
+      # @!attribute [r] old_identity
+      #   Optional attribute which keeps old diaspora* ID. Must be present when author attribute contains new diaspora*
+      #   ID.
+      #   @return [String] old identity
+      property :old_identity, :string, default: nil
+
+      # Returns diaspora* ID of the old person identity.
+      # @return [String] diaspora* ID of the old person identity
+      def old_identity
+        return @old_identity if author_is_new_id?
         author
+      end
+
+      # Returns diaspora* ID of the new person identity.
+      # @return [String] diaspora* ID of the new person identity
+      def new_identity
+        profile.author
       end
 
       # @return [String] string representation of this object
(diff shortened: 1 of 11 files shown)


spec/lib/diaspora_federation/entities/account_migration_signer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    it_behaves_like "a signable" do
+      let(:test_class) { TestAMSignableEntity }
+      let(:test_string) { "AccountMigration:1:2" }
+    end
+
+    describe "#unique_migration_descriptor" do
+      it "composes a string using #old_user_id and #new_user_id" do
+        expect(entity).to receive(:old_user_id).and_return("old")
```
SuperTux88: I don't think this needs mocking here. I would just use "old" and "new" in the TestAMSignableEntity above and then test that this test and the test below both return "AccountMigration:old:new". I don't like when tests just are an exact copy of the method but with mocking every single method call. I did this myself in the past, but try now not do do it. The problem with this kind of tests is, that you need to change the tests everytime you change something at the code (because they are an exact copy). For example the `#signature_data` test below, you would need to change it, when you rename this `#unique_migration_descriptor` method, even this doesn't change anything with the logic.

So these line by line, method call by method call tests are more complex, take more time to write (you just write the same code again), and make refactorings harder. So I think a little bit less is more in this case.

spec/lib/diaspora_federation/entities/account_migration_spec.rb
```diff
@@ -1,21 +1,110 @@
...
+
     let(:data) {
       hash.tap {|hash|
         properties = described_class.new(hash).send(:enriched_properties)
         hash[:signature] = properties[:signature]
       }
     }
-    let(:signature_data) { "AccountMigration:#{hash[:author]}:#{new_diaspora_id}" }
+    let(:signature_data) { "AccountMigration:#{old_diaspora_id}:#{new_diaspora_id}" }
+    alias_method :string, :signature_data
```
SuperTux88: Personally I would just write `let(:string) { signature_data }`, because I think it looks nicer ;) But whatever, this works for me if you like this more :)

spec/lib/diaspora_federation/entities/account_migration_spec.rb
```diff
@@ -1,21 +1,110 @@
...
+          expect_callback(:fetch_public_key, signer_id).and_return(signer_pkey.public_key)
+
+          expect {
+            Entities::AccountMigration.from_hash(hash)
+          }.to raise_error Entities::AccountMigration::SignatureVerificationFailed
+        end
+      end
+    end
+
+    context "with base message format" do
```
SuperTux88: "with old identity as author"

spec/lib/diaspora_federation/entities/account_migration_spec.rb
```diff
@@ -37,82 +126,45 @@ module DiasporaFederation
...
-        expect {
-          Entities::AccountMigration.new(hash).verify_signature
-        }.to raise_error Entities::AccountMigration::PublicKeyNotFound
-      end
-
-      it "raises when bad author signature was passed" do
-        hash[:signature] = "abcdef"
-
-        expect_callback(:fetch_public_key, new_diaspora_id).and_return(new_author_pkey.public_key)
+    context "with alternative message format" do
```
SuperTux88: "with new identity as author"

lib/diaspora_federation/entities/account_migration.rb
```diff
@@ -23,16 +25,25 @@ class AccountMigration < Entity
...
-      def to_s
-        "AccountMigration:#{author}:#{profile.author}"
+      # @!attribute [r] old_user
+      #   Optional attribute which is used to indicate old diaspora* ID when author attribute contains new diaspora* ID
+      #   @return [String] old_user
+      property :old_user, :string, optional: true
+
+      # Returns diaspora* ID of the old person depending on the format of the message
+      # @return [String] diaspora* ID of the old person
+      def old_user_id
```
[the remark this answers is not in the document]
cmrd-senya: Currently our property accessors in the entity classes have two responsibilities:
- they act as public API methods for the application to get entity properties
- they act as data source for serialization routines

In some cases, like this one, it may be useful to have a possibility to implement different behaviors for these two usages.

Currently `enriched_properties` provides the ability to change the serialized data in arbitrary way. And therefore we can change the way properties are serialized. However there are some problems with this approach. First it is not obvious that "enrich" is something that replaces the parameter. It looks hackish that we "enrich" some value by nullifying it. Also this way we have to store the unchanged value and only afterwards apply some changes instead of directly reading a correct value for serialization in the first place. And this feels like a more complex behavior, it's harder to follow.

What we can do instead is create a separate method for the serialization responsibility and make a call to respective public API method a default behavior. It can be easily achieved like this:
```ruby
diff --git a/lib/diaspora_federation/entity.rb b/lib/diaspora_federation/entity.rb
index 30549b3..d8c010f 100644
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -264,10 +264,14 @@ module DiasporaFederation
     # @return [Hash] hash with all properties
     def properties
       self.class.class_props.keys.each_with_object({}) do |prop, hash|
-        hash[prop] = public_send(prop)
+        hash[prop] = fetch_property_for_serialization(prop)
       end
     end
 
+    def fetch_property_for_serialization(prop)
+      public_send(prop)
+    end
+
     def normalized_properties
       properties.map {|name, value| [name, normalize_property(name, value)] }.to_h
     end

```

When it is done we can override the default behavior when we want so. Here is an example for the AM entity:
```ruby
      def fetch_property_for_serialization(prop)
        case prop
        when :old_identity
          @old_identity  
        else
          super
        end
      end
```

We can also use the same approach for signatures serialization instead of the current way we are doing it right now (using enriched properties). `enriched_properties` are still useful though, e.g. for adding the `addtional_data` to the serialization hash.

This would allow us to have an adequate public API method and also a straightforward serialization behavior. And therefore we can store the `old_identity` value only when it is mandatory.

Also there are some other possible ways how we can implement the override of the serialization getter. E.g. we can make some `serialization_accessor:` parameter for properties DSL or checking for some specific names for serialization getter, like searching for `#{prop}_for_serialization` method to fetch parrameter value and fall back to `public_send(prop)` in case it is missing. In this comment I just showed the simpliest possible implementation.

At the same time, I think that allowing `old_identity` to be there always even when it is not required is okay. And I agree that relying on `author` to decide which keys to sign/verify signature with is a better idea.

What do you think about this?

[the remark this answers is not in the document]
SuperTux88: Yes, you're right. Setting something to `nil` isn't really "enrich" ;) But I think for everything else we currently use `enriched_properties` (signatures for example), it's OK.

While adding a method for serialization could have some advantaged, I think for now it would only add another layer between the properties and serialization. Another place where something could be changed. So I'm not sure if it's worth the additional complexity, when we don't need it.

If we need it, I would probably prefer a special method name to get a property for serialization, because it looks the easiest to me.

But if we agree that we allow `old_identity` always, we don't need to change `enriched_properties` now, so this would be out of scope for this PR. So we can do this refactoring in another PR, or just wait until we really need this additional layer.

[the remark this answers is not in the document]
cmrd-senya: I see that you like the solution with less possible changes and I'm okay with that :) We can always add this feature later if we have stronger reasons for this.

cmrd-senya:
Updated

lib/diaspora_federation/entities/account_migration.rb
```diff
@@ -19,20 +21,39 @@ class AccountMigration < Entity
...
+      # @!attribute [r] old_identity
+      #   Optional attribute which keeps old diaspora* ID. Must be present when author attribute contains new diaspora*
+      #   ID.
+      #   @return [String] old identity
+      property :old_identity, :string, optional: true
+
+      # Returns diaspora* ID of the old person identity.
+      # @return [String] diaspora* ID of the old person identity
+      def old_identity
+        return old_identity_prop if author_is_new_id?
```
SuperTux88: 👍 (I would have missed that, but `@old identity || author` could be dangerous, when author + new indentity  is used for the signatures (because they aren't equal), but `old_identity` returns  something different)

lib/diaspora_federation/entities/account_migration.rb
```diff
@@ -44,30 +65,44 @@ def self.from_hash(*args)
 
       private
 
-      # @see DiasporaFederation::Entities::Signable#signature_data
-      def signature_data
-        to_s
+      def old_identity_prop
```
SuperTux88: I don't know if we need this, this is private anyway, so I would just use `@old_identity` directly internally.

lib/diaspora_federation/entities/account_migration.rb
```diff
@@ -44,30 +65,44 @@ def self.from_hash(*args)
...
+        @old_identity
+      end
+
+      def validate
```
SuperTux88: I would just use `rule :old_identity, :not_empty` in the normal validator, which would work, since we always return something, even when there wasn't anything in the hash or received xml. We only return nothing if the author is the new identity and the old should be set here, so it works as we need it. We would need to replace `optional: true` with `default: nil` (which I think is OK, since it's only half-optional anyway).

cmrd-senya: Maybe we can validate it with `diaspora_id` rule then?
```ruby
rule :old_identity, :diaspora_id
```

SuperTux88: Oh yeah, of course, that makes sense. :+1:

cmrd-senya:
Updated

SuperTux88:
Thank you, merged 🍪 (I changed one little thing after merging it with the spec, that I didn't see yesterday, but didn't want to do another round for this)

SuperTux88:
Oh, pushed it to the wrong repo, now it's merged ...

PR #90, opened 2017-11-05
Fix when booleans in relayables are false

SuperTux88:
This was a problem with dislikes.

It's safe to use `to_s` here, because we have only strings, numbers and booleans here anyway, because relayables don't support nested entities. `to_s` is used to generate the `signature_data` string too.

https://github.com/diaspora/diaspora_federation/blob/7503e9a804ea2ccf27ad7edb70da42b6870a548f/lib/diaspora_federation/entities/relayable.rb#L161-L164


cmrd-senya:
Merged, thanks :cookie:

PR #95, opened 2018-01-25
Fix relayable signatures for messages with invalid XML characters

SuperTux88:
Sometimes messages contain characters that are invalid for XML, but they are filteres out before creating the XML, otherwise the property would be empty in the XML.

But for relayables the value is also used for creating the signatures, so we need to filter the invalid characters earlier, before calculating the signature.

(This time @deusfigendi broke it with [this comment](https://pod.geraspora.de/posts/7493004#ecfcb7b0e35201351c81101b0e8ace24), which didn't federate. Thanks for breaking it 🍪 )

### DIFF
--- a/lib/diaspora_federation/entity.rb
+++ b/lib/diaspora_federation/entity.rb
@@ -277,7 +277,7 @@ module DiasporaFederation
 
       case self.class.class_props[name]
       when :string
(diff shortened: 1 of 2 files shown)


DeusFigendi:
You're welcome!

PR #97, opened 2018-02-14
Fix optional aware validator

SuperTux88:
Get optional props directly from the validated object which is much easier and safer than "guessing" the class name based on the validator name. That can cause a problem when another class with the same name is found (diaspora depends on the `webfinger` gem which has a `WebFinger` module, which obviously doesn't have the `optional_props` method). The "guessing" was only added because we used `OpenStruct` in the tests, but we shouldn't change the code only to make tests run. I changed the tests to use the real entities, with auto-validation disabled in the constructor, so we can test the validator manually.

Using the real entities for the tests also uncovered some bugs where for example empty strings are converted to nil and the validation wasn't invalid in this case, but should be.

This is a regression of #76

### DIFF
--- a/lib/diaspora_federation/discovery/h_card.rb
+++ b/lib/diaspora_federation/discovery/h_card.rb
@@ -48,7 +48,7 @@ module DiasporaFederation
       # @!attribute [r] nickname
       #   The first part of the diaspora* ID
       #   @return [String] nickname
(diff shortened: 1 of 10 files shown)


SuperTux88:
Merged, thanks for the review @jhass 🍪

PR #102, opened 2018-04-15
Update federation_entities.json

cmrd-senya:
parent_type is mandatory for likes

ref https://github.com/diaspora/diaspora/pull/7660/files/bf3f51ed496ef1bc0477cf346221c8926d6dde48

### DIFF
--- a/lib/diaspora_federation/schemas/federation_entities.json
+++ b/lib/diaspora_federation/schemas/federation_entities.json
@@ -101,7 +101,7 @@
                 "positive": { "type": "boolean" },
                 "parent_type": { "enum": ["Post", "Comment"] }
               },
-              "required": ["positive"]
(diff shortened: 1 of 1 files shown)


SuperTux88:
In the future, please don't create branches on the main repo (because when somebody forks the repo then, these temporary branches are forked too). I know this happens when you use the GitHub web editor to edit a file and you have write permissions, so it was probably by accident.

Anyway, thanks for the PR :cookie:

cmrd-senya:
> In the future, please don't create branches on the main  repo (because when somebody forks the repo then, these temporary  branches are forked too). I know this happens when you use the GitHub  web editor to edit a file and you have write permissions, so it was  probably by accident.

I'm not sure there was an option to create the branch on a fork, maybe I just didn't notice it.

PR #110, opened 2019-04-28
Entities JSON schema: width & height are optional

cmrd-senya:


SuperTux88:
Merged and released as 0.2.6

PR #113, opened 2019-10-21
Update dependencies and remove old versions

SuperTux88:
* Remove rails 4, and update to rails 6
* Remove old ruby versions, and add 2.6
* Update bundler
* Update all gems


.rubocop.yml
```diff
@@ -1,5 +1,7 @@
+require: rubocop-rails
+
 AllCops:
-  TargetRubyVersion: 2.1
+  TargetRubyVersion: 2.5
```
jhass: 2.6?

.travis.yml
```diff
@@ -1,21 +1,12 @@
 language: ruby
 
 rvm:
-  - 2.5.5
-  - 2.4.5
-  - 2.3.8
-  - 2.2.10
-  - 2.1
+  - 2.6.5
```
denschub: Do we have to specify exact versions here, or can we declare `2.6` instead? I don't think it's needed to pin an exact version, and there might be some gain if we automatically slide into new hotfix versions (in case something ever happens to break)

SuperTux88: With older ruby versions only 2.1 worked, but with newer versions this broke, that's why 2.2+ have the exact patch version (we also have the same problem in diaspora itself). :(

denschub: Okay, that's fine. Thanks for checking. :)

PR #114, opened 2019-10-30
Remove old federation

SuperTux88:
This removes all code that was used by the old federation and was still needed to receive messages from old diaspora pods. Since 0.7.0.0 diaspora only sends the new messages, this was over two years ago, so it should be safe to remove it.


SuperTux88:
Self-Reviewed after over 1.5 years, found that I forgot to remove legacy helpers for removed specs, and did some smaller cleanups in the comments integration test. So I guess this is safe to merge now.

Flaburgan:
Thank you for your work Benjamin, I wrote a small message about it to put visibility on your work. web+diaspora://fla@diaspora-fr.org/post/005f01b0bba50139ec5132a01d0dfba2

PR #116, opened 2021-06-16
Update faraday and faraday_middleware to 1.0

### LINKED ISSUE #115: Update faraday dependency to 1.0
Faraday is splitting its http adapters to separate gems and targets 2.0 for this. Should we pick a particular adapter here like faraday-net_http which uses net_http from standard ruby instead of faraday ?

pravi:
Closes #115 

1 test is failing

```

  1) DiasporaFederation::HttpClient.get follows redirects not more than 4 times
     Failure/Error: expect { HttpClient.get("http://www.example.com") }.to raise_error FaradayMiddleware::RedirectLimitReached

     NameError:
       uninitialized constant FaradayMiddleware::RedirectLimitReached
     # ./spec/lib/diaspora_federation/http_client_spec.rb:36:in `block (3 levels) in <module:DiasporaFederation>'
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,8 +2,8 @@ PATH
   remote: .
   specs:
     diaspora_federation (0.2.6)
-      faraday (>= 0.9.0, < 0.18.0)
-      faraday_middleware (>= 0.10.0, < 0.14.0)
+      faraday (~> 1.0)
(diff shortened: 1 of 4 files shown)


pravi:
@SuperTux88 thanks, can you make a new release so I can try updating Faraday in diaspora as well.

PR #121, opened 2021-11-27
Drop ruby 2.5 (EOL) and 2.6 (EOL soon) and update gems

SuperTux88:
Also enable MFA requirement for next version release.


Gemfile.lock
```diff
@@ -51,68 +51,74 @@ GEM
     diff-lcs (1.4.4)
     docile (1.4.0)
     erubi (1.10.0)
-    ethon (0.14.0)
+    ethon (0.15.0)
       ffi (>= 1.15.0)
     fabrication (2.22.0)
-    faraday (1.4.3)
+    faraday (1.8.0)
```
tclaus, line 57: For newer Rails Versions, shouldn't faraday be updated to a "~> 2" Version?

.rubocop.yml
```diff
@@ -1,7 +1,9 @@
-require: rubocop-rails
+require:
+  - rubocop-rails
+  - rubocop-rake
 
 AllCops:
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.6
```
tclaus, line 6: better 2.7+ ?

test/gemfiles/no-rails.Gemfile.lock
```diff
@@ -16,31 +16,37 @@ PATH
...
     crack (0.4.5)
       rexml
     diff-lcs (1.4.4)
     docile (1.4.0)
-    ethon (0.14.0)
+    ethon (0.15.0)
       ffi (>= 1.15.0)
     fabrication (2.22.0)
-    faraday (1.4.3)
+    faraday (1.8.0)
```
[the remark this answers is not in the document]
SuperTux88: Version 2 wasn't released yet when I created this PR. And since this gem already supports rails 6, it clearly works also with faraday 1, but I haven't looked into rails 7 yet and if there is a problem with rails 7 and faraday 1.x.

But I looked into if I should try to update to faraday 2 a few days ago and diaspora has other dependencies which still require faraday <2, so I would need to find a way to support both versions at the same time, because we can't use >=2 yet. But I haven't looked further into how easy it is to support both versions (because I thought faraday 2 has some breaking changes), and I'd probably prefer doing that in a separate PR.

.rubocop.yml
SuperTux88: Back when I created this PR, 2.6 was still fine, but it's getting close to also be EOL, so maybe we should also already drop it? :thinking:

denschub: > so maybe we should also already drop it? 

Yeah, 2.6 will be EOL at the end of this month, 2022-03-31 - so I don't think continuing to support it is reasonable.

SuperTux88: Ok, I dropped 2.6 and added 3.1 instead (and also fixed that 3.0 stays 3.0 for GH-Actions)

PR #122, opened 2021-11-27
Remove legacy WebFinger support/fallback

SuperTux88:
RFC 7033 WebFinger is supported since a long time (diaspora 0.7.1.0), so at this point this is basically only duplicating requests when users don't exists because it still retries with fallback.


denschub (approved):
This looks reasonable, and the test coverage seems perfect.

Issue #124, opened 2022-03-29
Comments dont not federate thread_parent_guid attribute as states in doc


tclaus:
In documentation the thread_parent_guid attribute is part of the comment entity.
Is this attribute really distributed? If not, this would be nice to see this in next release of federation protocol.

SuperTux88:
It isn't in the library yet because diaspora doesn't support threads in comments yet. It is in the protocol specification because Friendica is already using it. When it gets added to the library, it needs to be actively added to the diaspora database and handled everywhere, that's the reason why it wasn't added here yet. Now it implicitly just works with the `additional_data` (so federation between friendica and diaspora correctly handles the field and forwards it with a valid signature), but when it gets explicitly added to the library that stops working and it needs explicit handling and support from the diaspora side.

Since it's only a field, it can be added relatively quickly if needed.

SuperTux88:
> I have seen this attribute in a database.. Butt this just was a dream..

It is in the database, as data, on comments that are in a thread (comments received from friendica) in the `additional_data`.

> By the way: A comment in Diaspora already is a "Commentable" so it is prepared to comment a comment in terms of infrastructure.

A comment isn't a "Commentable", neither in diaspora, nor in the protocol specification. The only allowed comment parents are posts (status messages or reshares). If you comment on a comment you need to have the parent post still as parent, because only then the federation of the comment works, because only the post knows to where it was federated, and the comment to which you are responding goes into `thread_parent_guid`.

> Older clients then will not show these in its UI.

Old diaspora versions will just display the comments in chronological order without the thread information, like diaspora currently just shows friendica comments without the thread information, but still correctly federates the thread information to other servers (for example back to another friendica which then displays the thread information again).

> Of course, diaspora then needs a minimal change to show sub-comments, and a bigger change in UI and handling.

As said, the changes here in the library are easy and can be done quickly (only add the attribute at a few places), but the changes in diaspora to actually support the attribute are probably more work. So it's probably better to prepare the diaspora side first before adding the attribute here.

tclaus:
Now I know about "comments_signature" table and the additional_data there. Thanks for clarification.

tclaus:
If diaspora would support threaded comments - could these parent_guids simply written back to the additional_data and getting federation for free?

PR #125, opened 2022-03-29
Fixes 124 Adding missing attribute

tclaus:
Adds the thread_parent_guid_ attribute to comment as described in the docs.


lib/diaspora_federation/entities/comment.rb
```diff
@@ -24,6 +24,12 @@ class Comment < Entity
       #   The timestamp when the comment was edited
       #   @return [Time] edited time
       property :edited_at, :timestamp, optional: true
+
+      #@!attribute [r] guid
```
diaspora-code-review, line 28: Layout/LeadingCommentSpace: Missing space after `#`.

spec/lib/diaspora_federation/entities/comment_spec.rb
```diff
@@ -4,13 +4,15 @@ module DiasporaFederation
   describe Entities::Comment do
     let(:parent) { Fabricate(:post, author: bob) }
     let(:parent_entity) { Fabricate(:related_entity, author: bob.diaspora_id) }
-    let(:data) {
+    let(:parent_comment) { Fabricate(:comment)}
```
diaspora-code-review, line 7: Layout/SpaceInsideBlockBraces: Space missing inside }.

diaspora-code-review, line 7: Layout/TrailingWhitespace: Trailing whitespace detected.

spec/lib/diaspora_federation/entities/comment_spec.rb
```diff
@@ -4,13 +4,15 @@ module DiasporaFederation
   describe Entities::Comment do
     let(:parent) { Fabricate(:post, author: bob) }
     let(:parent_entity) { Fabricate(:related_entity, author: bob.diaspora_id) }
-    let(:data) {
+    let(:parent_comment) { Fabricate(:comment)} 
+    let(:data) {
```
diaspora-code-review, line 8: Layout/TrailingWhitespace: Trailing whitespace detected.

spec/lib/diaspora_federation/entities/comment_spec.rb
```diff
@@ -4,13 +4,15 @@ module DiasporaFederation
...
-    let(:data) {
+    let(:parent_comment) { Fabricate(:comment)} 
+    let(:data) { 
       Fabricate
         .attributes_for(
           :comment_entity,
           author:      alice.diaspora_id,
           parent_guid: parent.guid,
-          parent:      parent_entity
+          parent:      parent_entity,
```
diaspora-code-review, line 14: Layout/HashAlignment: Align the keys and values of a hash literal if they span more than one line.

lib/diaspora_federation/entities/comment.rb
```diff
@@ -24,6 +24,12 @@ class Comment < Entity
       #   The timestamp when the comment was edited
       #   @return [Time] edited time
       property :edited_at, :timestamp, optional: true
+
+      # @!attribute [r] guid
+      #   A random string of at least 16 chars pointing to a parent comment.
+      #   @see Validation::Rule::Guid
+      #   @return [String] parent comment guid
+      property :thread_parent_guid, :string, optional: true
```
SuperTux88, line 32: The field should also be added to the `CommentValidator` to validate it's a valid GUID.
And it should also be added to the comment entity in `lib/diaspora_federation/schemas/federation_entities.json`.

SuperTux88 (changes requested):
As already mentioned in https://github.com/diaspora/diaspora_federation/issues/124#issuecomment-1082468206, this isn't in the library yet intentionally, because adding it would require explicit handling on the diaspora side. So unless you also want to add support and handling for it on the diaspora side (which will be a lot more work than just adding the field here, since it also requires a migration of existing data from friendica), I wouldn't recommend to add this yet. Because if this gets merged we wouldn't be able to upgrade the lib anymore for diaspora until it's properly handled there.

Also sidenote: A more useful commit message would be appreciated where a short "what was done" is in the first line, and a `fixes #124` in an additional line so GitHub actually detects that and auto-closes the issue when it gets merged. Currently it's kinda the other way around, but without the `#` GitHub doesn't even detect the issue.

PR #130, opened 2023-06-15
[CI] Run Pronto on GitHub Actions.

denschub:
This is https://github.com/diaspora/diaspora/pull/8429, but instead of over there, it's here.


jhass (approved):

SuperTux88 (approved):
Thanks :cookie:

